PR #29, opened 2020-04-19
Fix not to crash when unsubscribing while publishing a message

styner32:
1. The app crashes when unsubscribing while publishing a message to the unsubscribed channel because it sends a message in goroutine. https://github.com/walletlink/walletlink/blob/2.0.2/server/rpc/pub_sub.go#L124 ~~The fix is to wait for all subscribers to get a message, so safely unlock https://github.com/walletlink/walletlink/blob/2.0.2/server/rpc/pub_sub.go#L114 to unsubscribe~~ Created a lock for each subscriber and using it when publishing and unsubscribing

2. Fix tests (https://github.com/walletlink/walletlink/pull/1/commits/c69a594e8a8666283e5fbec3092fcb2a2ac40ad5). Failed because of https://github.com/walletlink/walletlink/commit/19dda19c607bcb8dfcc39ebb22508f8843424716#diff-588e74a11db2e82785409f89372885bf

### DIFF
--- a/server/rpc/pub_sub.go
+++ b/server/rpc/pub_sub.go
@@ -17,16 +17,18 @@ type subscriberSet map[Subscriber]struct{}
 
 // PubSub - pub/sub interface for message senders
 type PubSub struct {
-	lock   sync.Mutex
-	subMap map[string]subscriberSet      // Subscription ID -> Subscribers
-	idMap  map[Subscriber]util.StringSet // Subscriber -> Subscription IDs
+	lock     sync.Mutex
+	subMap   map[string]subscriberSet      // Subscription ID -> Subscribers
+	idMap    map[Subscriber]util.StringSet // Subscriber -> Subscription IDs
+	subLocks map[Subscriber]*sync.Mutex    // Subscriber -> mutex lock
 }
 
 // NewPubSub - construct a PubSub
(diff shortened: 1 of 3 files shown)


petejkim:
Documenting IRL feedback: I think we should explore creating a unique lock for each subscriber and using that in publish and unsubscribe.

styner32:
Fixed as addressed

styner32:
> Documenting IRL feedback: I think we should explore creating a unique lock for each subscriber and using that in publish and unsubscribe.

Fixed as addressed

server/rpc/pub_sub.go
```diff
@@ -17,16 +17,18 @@ type subscriberSet map[Subscriber]struct{}
 
 // PubSub - pub/sub interface for message senders
 type PubSub struct {
-	lock   sync.Mutex
-	subMap map[string]subscriberSet      // Subscription ID -> Subscribers
-	idMap  map[Subscriber]util.StringSet // Subscriber -> Subscription IDs
+	lock     sync.Mutex
```
petejkim, line 20: Please also change this to *sync.Mutex since mutexes aren't supposed to be copied. It works if there's only one instance, but if it's moved around it will get copied

PR #33, opened 2020-05-15
Monitor active connections

styner32:
1. Add timeout to webhook
2. Add endpoints to monitor goroutines and heap memory
3. Monitor number of active connections and make it crash when there are too many open files
4. Set write deadline for writing messages from websocket connection

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -58,6 +58,9 @@ var (
 
 	// ReadDeadline - Deadline for reading messages
 	ReadDeadline time.Duration
+
+	// WriteDeadline - Deadline for writing messages
+	WriteDeadline time.Duration
 )
 
 func init() {
@@ -67,6 +70,9 @@ func init() {
 
 	readDeadlineSecs, _ := strconv.Atoi(getEnv("READ_DEADLINE_SECS", "30"))
 	ReadDeadline = time.Second * time.Duration(readDeadlineSecs)
+
+	writeDeadlineSecs, _ := strconv.Atoi(getEnv("WRITE_DEADLINE_SECS", "30"))
+	WriteDeadline = time.Second * time.Duration(writeDeadlineSecs)
 }
 
 func getEnv(name string, defaultValue string) string {
--- a/config/config.go
+++ b/config/config.go
@@ -61,6 +61,9 @@ var (
 
 	// WriteDeadline - Deadline for writing messages
 	WriteDeadline time.Duration
+
+	// For debugging
+	BasicAuthPassword = getEnv("BASIC_AUTH_PASSWORD", "")
 )
 
 func init() {
--- a/server/rpc.go
+++ b/server/rpc.go
@@ -55,10 +55,28 @@ func (srv *Server) rpcHandler(w http.ResponseWriter, r *http.Request) {
 			if !ok {
 				return
 			}
+
 			if v, ok := res.(rune); ok && v == 'h' {
-				ws.WriteMessage(websocket.TextMessage, []byte("h"))
+				if srv.writeDeadline > 0 {
+					if err := ws.SetWriteDeadline(time.Now().Add(srv.writeDeadline)); err != nil {
+						log.Println(errors.Wrap(err, "websocket set write deadline failed"))
+						break
+					}
+				}
+
+				if err := ws.WriteMessage(websocket.TextMessage, []byte("h")); err != nil {
+					log.Println(errors.Wrap(err, "websocket write heartbeat"))
+				}
 				continue
 			}
+
+			if srv.writeDeadline > 0 {
+				if err := ws.SetWriteDeadline(time.Now().Add(srv.writeDeadline)); err != nil {
+					log.Println(errors.Wrap(err, "websocket set write deadline failed"))
+					break
+				}
+			}
+
 			if err := ws.WriteJSON(res); err != nil {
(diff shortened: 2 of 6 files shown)


server/server.go
```diff
@@ -54,13 +69,16 @@ func NewServer(options *NewServerOptions) *Server {
...
-		router:         router,
-		store:          st,
-		pubSub:         rpc.NewPubSub(),
-		allowedOrigins: options.AllowedOrigins,
-		webhook:        options.Webhook,
-		serverURL:      options.ServerURL,
-		readDeadline:   options.ReadDeadline,
+		router:            router,
+		store:             st,
+		pubSub:            rpc.NewPubSub(),
```
hishbase, line 74: I'm not seeing the code that fixes the panic in `pub_sub.go` on line 142. With the current setup, WalletLink will crash if there's a race condition and we write to a close channel

image

config/config.go
```diff
@@ -67,6 +73,9 @@ func init() {
 
 	readDeadlineSecs, _ := strconv.Atoi(getEnv("READ_DEADLINE_SECS", "30"))
 	ReadDeadline = time.Second * time.Duration(readDeadlineSecs)
+
+	writeDeadlineSecs, _ := strconv.Atoi(getEnv("WRITE_DEADLINE_SECS", "30"))
```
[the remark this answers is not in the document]
styner32: I can do it, but this is configurable right now

server/server.go
```diff
@@ -38,8 +48,13 @@ type NewServerOptions struct {
 	ServerURL      string
 	ForceSSL       bool
 	ReadDeadline   time.Duration
+	WriteDeadline  time.Duration
 }
 
+const (
+	activeConnsThreshold = 5000
```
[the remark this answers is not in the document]
styner32: it was a temporal solution to restart the server.
I can make it configurable, but if it hits 2000, the server already stops responding
we should make this number low if we want to catch it earlier

server/server.go
styner32: I haven't seen any panics in `pub_sub.go` after this fix https://github.com/walletlink/walletlink/pull/29

The issue right now is the lock https://github.com/walletlink/walletlink/blob/master/server/rpc/pub_sub.go#L140 never get unlocked.

config/config.go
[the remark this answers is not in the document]
hishbase: Gotcha. What do you have it set to on Heroku? Can we bring it down to `15`? Also, lets move the default to 15 seconds. 30 seconds is an eternity

server/rpc.go
```diff
@@ -31,31 +36,56 @@ func (srv *Server) rpcHandler(w http.ResponseWriter, r *http.Request) {
...
 		log.Println(errors.Wrap(err, "websocket upgrade failed"))
 		return
 	}
-	defer ws.Close()
+
+	log.Printf("Opened ws with IP: %s, %s User-Agent: %s", clientIP(r), r.RemoteAddr, r.Header.Get("User-Agent"))
+	defer func() {
+		log.Printf("Closed ws with IP: %s, %s User-Agent: %s", clientIP(r), r.RemoteAddr, r.Header.Get("User-Agent"))
+		ws.Close()
+	}()
```
hishbase, line 48: please keep things at 80 character limit. Do this everywhere

server/server.go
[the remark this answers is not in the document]
hishbase: gotcha

server/server.go
hishbase: We probably haven't seen it because of the lock issue you mentioned but once the lock issue is resolved this can crash the server i.e. we will be writing to a closed channel.

Can we fix it as part of this PR?

server/rpc.go
```diff
@@ -31,31 +36,56 @@ func (srv *Server) rpcHandler(w http.ResponseWriter, r *http.Request) {
...
+
+			if srv.writeDeadline > 0 {
+				ws.SetWriteDeadline(time.Now().Add(srv.writeDeadline))
+			}
+
 			if err := ws.WriteJSON(res); err != nil {
 				log.Println(errors.Wrap(err, "websocket write failed"))
 				ws.Close()
-				return
+				isClosed = true
```
hishbase, line 88: any reason we're setting the flag?Why not just log here and exit the goroutine? instead of having it to continue to for-loop?

hishbase: Also, I'm not sure this is fixing anything? Isn't it going to hang the goroutine?

server/server.go
styner32: As far as I know there is no way to check if a channel is closed in go. Do you any suggestions to prevent the app from crashing?

server/rpc.go
styner32: ah if we just return here, `sendCh` will stay opened for reading and it will make https://github.com/walletlink/walletlink/blob/master/server/rpc/pub_sub.go#L143 blocked.

So the solution is reading messages from the channel until it get closed.

config/config.go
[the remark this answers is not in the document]
styner32: yes. it is set to 15 seconds for both reading and writing

styner32:
Closing it and continuing in https://github.com/walletlink/walletlink/pull/34

PR #34, opened 2020-05-25
Fix "too many connections" error

styner32:
1. Set timeout for webhook. This could leak file descriptors
2. Wait for a channel to be closed in a goroutine for handling publishing messages. This is to prevent blocking at https://github.com/walletlink/walletlink/blob/2.0.2/server/rpc/pub_sub.go#L124
3. Don't publish a message to an unsubscribed channel.
4. Configure a PG connection pool. This could open a lot of connections when db is too slow

### DIFF
--- a/cmd/walletlinkd/walletlinkd.go
+++ b/cmd/walletlinkd/walletlinkd.go
@@ -32,7 +32,14 @@ func main() {
 		st = store.NewMemoryStore()
 	} else {
 		var err error
-		st, err = store.NewPostgresStore(config.PostgresURL, "store")
+		st, err = store.NewPostgresStore(
+			config.PostgresURL,
+			"store",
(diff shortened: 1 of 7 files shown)


kevin71089:
> 1. Set timeout for webhook. This could leak file descriptors
> 2. Wait for a channel to be closed in a goroutine for handling publishing messages. This is to prevent blocking at https://github.com/walletlink/walletlink/blob/2.0.2/server/rpc/pub_sub.go#L124
> 3. Don't publish a message to an unsubscribed channel.
> 4. Configure a PG connection pool. This could open a lot of connections when db is too slow

pepe-invest-git (changes requested):
GROUP-BANK

mrcali213 (approved):

bangtoven:
thanks

PR #61, opened 2021-01-18
Sign typed data calls receive a stringified version of the json message

alexshoykhet:
Per eth-sig-util (and verifying with Metamask), the sign typed data calls receive a stringified version of the input json message.

### DIFF
--- a/js/src/provider/WalletLinkProvider.ts
+++ b/js/src/provider/WalletLinkProvider.ts
@@ -15,7 +15,8 @@ import {
   ensureBuffer,
   ensureHexString,
   ensureIntNumber,
-  ensureRegExpString
+  ensureRegExpString,
+  ensureString
 } from "../util"
 import eip712 from "../vendor-js/eth-eip712-util"
 import { FilterPolyfill } from "./FilterPolyfill"
@@ -650,15 +651,15 @@ export class WalletLinkProvider implements Web3Provider {
     params: unknown[]
   ): Promise<JSONRPCResponse> {
     this._requireAuthorization()
-    const typedData = params[0]
+    const typedDataStr = ensureString(params[0])
     const address = ensureAddressString(params[1])
 
     this._ensureKnownAddress(address)
 
-    const message = eip712.hashForSignTypedDataLegacy({ data: typedData })
-    const typedDataJson = JSON.stringify(typedData, null, 2)
+    const typedDataJson = JSON.parse(typedDataStr)
+    const message = eip712.hashForSignTypedDataLegacy({ data: typedDataJson })
 
-    return this._signEthereumMessage(message, address, false, typedDataJson)
+    return this._signEthereumMessage(message, address, false, typedDataStr)
   }
 
   private async _eth_signTypedData_v3(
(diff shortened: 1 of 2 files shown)


petejkim:
Might wanna check the type and support both.

alexshoykhet:
> I think we should check the type and support both.

Thanks! Done d8f96b47c91b5101c2bd8df0b756dab3ef4207d9

js/src/util.ts
```diff
@@ -151,6 +151,18 @@ export function ensureBN(val: unknown): BN {
...
+    return JSON.parse(val)
+  }
+
+  if (typeof val == "object") {
+    return val as Object
+  }
+
+  throw new Error(`Not a string or a json object: ${val}`)
+}
+
```
petejkim, line 165: I would suggest something like:

```
export function ensureParsedJsonObject<T extends object = any>(
  val: unknown
): T {
  if (typeof val === "string") {
    return JSON.parse(val)
  }

  if (typeof val === "object") {
    return val as T
  }

  throw new Error(`Not a JSON or an object: ${val}`)
}
```

1. JSON implies text, so `ensureJsonOr` may mislead the developer into thinking that it might return a string
2. We shouldn't use uppercase `Object`, instead a lowercase `object` but make it possible for the user to define the shape of the object if needed or default to any.
3. Always use `===`.

js/src/provider/WalletLinkProvider.ts
```diff
@@ -650,45 +651,45 @@ export class WalletLinkProvider implements Web3Provider {
     params: unknown[]
   ): Promise<JSONRPCResponse> {
     this._requireAuthorization()
-    const typedData = params[0]
+    const typedDataJson = ensureJsonOrParseString(params[0])
```
petejkim, line 654: call the variable `typedData` since it will always be an `object` rather than a JSON string?

js/src/provider/WalletLinkProvider.ts
```diff
@@ -650,45 +651,45 @@ export class WalletLinkProvider implements Web3Provider {
...
     const address = ensureAddressString(params[1])
 
     this._ensureKnownAddress(address)
 
-    const message = eip712.hashForSignTypedDataLegacy({ data: typedData })
-    const typedDataJson = JSON.stringify(typedData, null, 2)
+    const message = eip712.hashForSignTypedDataLegacy({ data: typedDataJson })
+    const typedDataStr = JSON.stringify(typedDataJson, null, 2)
```
petejkim, line 660: and this can be called `typedDataJson`

js/src/util.ts
alexshoykhet: Thanks, done c5e0d171e4a6fa75f11ce18f934709047f095034

js/src/provider/WalletLinkProvider.ts
alexshoykhet: c5e0d171e4a6fa75f11ce18f934709047f095034

js/src/provider/WalletLinkProvider.ts
alexshoykhet: c5e0d171e4a6fa75f11ce18f934709047f095034

alexshoykhet:
Respond to nit on Json vs JSON (we prefer JSON) - 5e746514b00cc4d5f68680e7a49641599bc0a632

petejkim:
The changes look good to me, and worked fine in tests. Could you squash and merge?

zaeguerra (approved):

zaeguerra (approved):

zaeguerra (approved):

PR #62, opened 2021-02-08
Implement EIP1193 Provider Methods

alexshoykhet:
Implement `request` and `on` per `EIP1193` spec. 

@petejkim one change I made here is to use `es2016` so we can import the `safe-emitter` library.  Thoughts on whether this is ok to do?

### DIFF
--- a/js/package.json
+++ b/js/package.json
@@ -41,7 +41,9 @@
     "bn.js": "^5.1.1",
     "clsx": "^1.1.0",
     "preact": "^10.5.9",
-    "rxjs": "^6.6.3"
+    "rxjs": "^6.6.3",
+    "eth-rpc-errors": "4.0.2",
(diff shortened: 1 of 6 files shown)


petejkim:
One problem that I see here is that by implementing these methods, it might make it seem that it supports subscriptions `.on("message")` but it doesn't. That breaks certain apps. 

This commit essentially reverts this commit:
https://github.com/walletlink/walletlink/commit/f7a62fb366b1ca7374c80bf68f3924bbfa48d997#diff-6a46db475c5a2259fb4fdda1453652bd0d687a775d9fa3a943f9f7cdda34f030

Which was removed because it was breaking feature detection.

To implement this correctly, all event emission must be implemented as well, or else there will be other side effects.

vishnumad:
PR https://github.com/walletlink/walletlink/pull/64 to add support for subscriptions

PR #64, opened 2021-02-26
Add EIP-1193 support

vishnumad:
Adds support for EIP-1193 provider methods

Also adds support for subscribing/unsubscribing to `newHeads` and `logs` events. `newPendingTransactions` and `syncing` are not supported.

Subscribe to a new event like so
```typescript
provider.request({ method: "eth_subscribe" params: ["newHeads"]})
```

The provider then emits new `message` events.
```typescript
provider.on("message", (data: ProviderMessage) => {})
```

Unsubscribe like so
```typescript
provider.request({ method: "eth_unsubscribe" params: ["<subscription id>"]})
```

### DIFF
--- a/js/src/provider/SubscriptionManager.ts
+++ b/js/src/provider/SubscriptionManager.ts
@@ -8,6 +8,18 @@ const PollingBlockTracker = require("eth-block-tracker")
 const createSubscriptionManager = require("eth-json-rpc-filters/subscriptionManager")
 const noop = () => {}
 
+export interface SubscriptionResult {
+  result?: unknown
+}
+
+export interface SubscriptionNotification {
+  method: string
+  params: {
+    subscription: string
+    result: unknown
+  }
+}
+
 export class SubscriptionManager {
   private readonly subscriptionMiddleware: SubscriptionMiddleware
   readonly events: SafeEventEmitter
@@ -41,8 +53,6 @@ export class SubscriptionManager {
     this.subscriptionMiddleware.destroy()
   }
 }
-
-type SubscriptionResult = { result?: any }
 
 interface SubscriptionMiddleware {
   (
--- a/js/src/provider/SubscriptionManager.ts
+++ b/js/src/provider/SubscriptionManager.ts
@@ -1,0 +1,36 @@
+import SafeEventEmitter from "@metamask/safe-event-emitter"
+import { Web3Provider } from "./Web3Provider"
+const PollingBlockTracker = require("eth-block-tracker")
+const createSubscriptionManager = require("eth-json-rpc-filters/subscriptionManager")
+const noop = () => {}
+
+export class SubscriptionManager {
+  private readonly subscriptionMiddleware: any
+  readonly events: SafeEventEmitter
+
+  constructor(provider: Web3Provider) {
+    const blockTracker = new PollingBlockTracker({
+      provider: provider,
+      pollingInterval: 15 * 1000, // 15 sec
+      setSkipCacheFlag: true
+    })
+
+    const { events, middleware } = createSubscriptionManager({
+      blockTracker: blockTracker,
+      provider: provider
+    })
+
+    this.events = events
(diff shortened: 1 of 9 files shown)


js/src/provider/SubscriptionManager.ts
```diff
@@ -0,0 +1,36 @@
+import SafeEventEmitter from "@metamask/safe-event-emitter"
+import { Web3Provider } from "./Web3Provider"
+const PollingBlockTracker = require("eth-block-tracker")
+const createSubscriptionManager = require("eth-json-rpc-filters/subscriptionManager")
+const noop = () => {}
+
+export class SubscriptionManager {
+  private readonly subscriptionMiddleware: any
```
petejkim, line 8: avoid the use of any here

js/src/provider/WalletLinkProvider.ts
```diff
@@ -58,18 +68,23 @@ export class WalletLinkProvider extends SafeEventEmitter implements Web3Provider
     const chainIdStr = prepend0x(this._chainId.toString(16))
 
     // indicate that we've connected, for EIP-1193 compliance
-    this.emit('connect', { chainIdStr });
+    this.emit("connect", { chainIdStr })
```
petejkim, line 71: Use the same prettier rule autoformatting rules used by this project.

petejkim:
Also can you submit a new PR against master please?

gigilove2001 (approved):

petejkim:
👀

js/src/provider/SubscriptionManager.ts
```diff
@@ -0,0 +1,66 @@
...
+
+export interface SubscriptionNotification {
+  method: string
+  params: {
+    subscription: string
+    result: unknown
+  }
+}
+
+export class SubscriptionManager {
```
brendanww, line 23: Please add class-level doc -- might be worth linking to https://eips.ethereum.org/EIPS/eip-1193

js/src/relay/WalletLinkRelay.ts
```diff
@@ -68,6 +68,7 @@ export interface WalletLinkRelayOptions {
 }
 
 export class WalletLinkRelay {
```
brendanww, line 70: Please add class-level doc

js/src/relay/WalletLinkRelay.ts
```diff
@@ -352,6 +353,20 @@ export class WalletLinkRelay {
...
 
+  public static makeRequestId(): number {
+    // max nextId == max int32 for compatibility with mobile
+    this._nextRequestId = (this._nextRequestId + 1) % 0x7fffffff
+    const id = this._nextRequestId
+    const idStr = prepend0x(id.toString(16))
+    // unlikely that this will ever be an issue, but just to be safe
+    const callback = WalletLinkRelay.callbacks.get(idStr)
+    if (callback) {
+      // TODO - how to handle this case
```
brendanww, line 364: Wouldn't it be safer to just try generating a new `id`?

If I understood the docs correctly, the id is for keeping track of requests and responses when sent over a websocket, so the consequence here would be mismatching a response to a request. No?

brendanww (approved):

alexshoykhet (approved):
Tested locally with web3 dapp.

js/src/relay/WalletLinkRelay.ts
petejkim: This code is taken from https://github.com/CoinbaseWallet/cipher-web-view/blob/50dd82456575dccda41f1f4629372096b6969d92/src/CipherProvider.ts#L133

ID is sequential, and once it reaches max int32, which will probably never happen, it will start from ID 0 again. When that happens, it's probably safe to say that the callback for ID 0 has already been called. (since 2 billion+ other requests have been called since then)

petejkim (approved):

zaeguerra (approved):

zaeguerra (approved):

PR #69, opened 2021-03-10
Replace existing provider errors with errors from eth-rpc-errors

vishnumad:
Remove the existing `ProviderError` class and replace the errors with ones from the `eth-rpc-errors` library


jasoncoker925:
vishnumad:vishnu/update-errors

PR #213, opened 2021-11-19
Add 3085 support for non-whitelisted networks

brendanww:
Sends addEthereumChain requests to clients for non-whitelisted networks instead of default rejecting the request.


js/src/provider/WalletLinkProvider.ts
```diff
@@ -940,16 +970,32 @@ export class WalletLinkProvider
...
 
+    if (request.rpcUrls?.length === 0) {
+      return { jsonrpc: '2.0', id: 0, error: { code: 2, message: `please pass in at least 1 rpcUrl` } };
+    }
+
     const chainIdNumber = parseInt(request.chainId, 16);
     const ethereumChain = EthereumChain.fromChainId(BigInt(chainIdNumber));
     if (ethereumChain === undefined) {
-      return { jsonrpc: '2.0', id: 0, error: { code: 2, message: `chainId ${request.chainId} not supported` } };
+      const success = await this.addEthereumChain(
```
hieronymus777, line 980: If the caller:
1. Adds new ethereum chain that isn't hardcoded in `EthereumChain.fromChainId`
2. Calls `wallet_switchEthereumChain`
The code will throw an error since `wallet_switchEthereumChain` will try to use `EthereumChain.fromChainId` on the unsupported chain.

Just want to clarify if this is the expected behavior or whether we should add support for it.

hieronymus777: For consistency, it looks like other wallets support calling `wallet_switchEthereumChain` on a newly added chain.

db4rTo: []()_**~~~~**_

js/src/provider/WalletLinkProvider.ts
```diff
@@ -214,16 +214,69 @@ export class WalletLinkProvider
     }
   }
 
-  private async switchEthereumChain(rpcUrl: string, chainId: number) {
+  private async addEthereumChain(
+    chainId: number,
```
amhedcb, line 218: What's the breakoff point when we change constructor to be objects instead of params?

js/src/provider/WalletLinkProvider.ts
```diff
@@ -214,16 +214,69 @@ export class WalletLinkProvider
...
+
+    if (typeof res.result === 'boolean') {
+      // legacy handling. to be deprecated in february 2022
+      if (res.result === true) {
+        this._storage.setItem(HAS_CHAIN_BEEN_SWITCHED_KEY, "true")
+        this.updateProviderInfo(rpcUrls[0], chainId, false)
+      }
+
+      return res.result === true
+    } else {
```
amhedcb, line 247: nit: can remove this `else` since you always return on line 246

js/src/provider/WalletLinkProvider.ts
```diff
@@ -966,32 +1019,31 @@ export class WalletLinkProvider
...
+    const success = await this.addEthereumChain(
+      chainIdNumber,
+      request.rpcUrls ?? [],
+      request.blockExplorerUrls,
+      request.chainName,
+      request.iconUrls,
+      request.nativeCurrency
+    )
+    if (success) {
+      return {jsonrpc: '2.0', id: 0, result: null};
```
amhedcb, line 1036: id is always 0?

spencerstock (approved):

js/src/provider/WalletLinkProvider.ts
brendanww: seems like something eslint should tell us, no?

PR #222, opened 2021-11-25
Generic Messages over WalletLink

spencerstock:
TDD is in progress for this idea, and this PR can serve as a proof-of-concept until we reach alignment on the approach used here. 

We're adding this method so that 2 clients with an established walletlink connection can send messages across walletlink without walletlink needing to parse parameters on the message. This would be an optimization over the approach where we add a function type for each call.

1. Sending clients sends a message over walletlink conforming to whatever key:value parameters it needs
2. Receiving client is aware of the message type, in code referred to as "action" and can parse out the parameters.
3. Receiving client then sends back an equally ambiguous message that passes through walletlink in a similar way.
4. Sending client receives and parses the response object

This approach provides extra security (no proprietary messages are exposed in public repo) and convenience (we don't require a new channel through logical layers for every new message type, this convenience is expanded on in our mobile clients as well) at the cost of type enforcement in the walletlink layer.

There was already an ArbitraryMessage type coded into walletLink that can accomplish this same goal, the type is iterated on here in order to smooth parsing on sending and receiving clients

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 .DS_Store
 *.test
 build/
+.idea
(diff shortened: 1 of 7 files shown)


Elrolon:
Approved

PR #227, opened 2021-12-02
Update json rpc url and chain id at the same time

hieronymus777:
Make provider listen for chain id changed event and update its chain accordingly.


heangpov16:
Support money

brendanww (approved):

allen-at-work (approved):

PR #247, opened 2021-12-15
Add override for `isCoinbaseWallet` in provider options

cb-jake:
## Description

Add override to disable isCoinbaseWallet in walletlink provider options

### DIFF
--- a/js/src/provider/WalletLinkProvider.ts
+++ b/js/src/provider/WalletLinkProvider.ts
@@ -53,7 +53,7 @@ export class WalletLinkProvider
   extends SafeEventEmitter
   implements Web3Provider {
   // So dapps can easily identify Coinbase Wallet for enabling features like 3085 network switcher menus
-  public readonly isCoinbaseWallet: boolean = true
+  public readonly isCoinbaseWallet: boolean
 
   private readonly _filterPolyfill = new FilterPolyfill(this)
   private readonly _subscriptionManager = new SubscriptionManager(this)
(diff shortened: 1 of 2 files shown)


erin-at-work (approved):

js/src/provider/WalletLinkProvider.ts
```diff
@@ -94,6 +95,10 @@ export class WalletLinkProvider
       ? options.walletLinkAnalytics
       : new WalletLinkAnalytics()
 
+    if (options.overrideIsCoinbaseWallet) {
```
hieronymus777, line 98: I think you need to check `if (typeof options.overrideIsCoinbaseWallet === 'boolean')`.

Otherwise `isCoinbaseWallet` will always be true since `overrideIsCoinbaseWallet = false` is falsey and won't set the `isCoinbaseWallet` variable to true.

cb-jake: Great catch. I'll set the default in the constructor to avoid this.

hieronymus777 (approved):

ShortyV21:
I know what GitHub is building, wireless free from restraints Humans beings so the world is forced to stop lying to them. The creator created our energies and our inner energy is most poweful

ShortyV21:
And they’re way of doing it is knowledge from the internet

ShortyV21:
I still remember being approached through the internet twice but I knew way before then

PR #263, opened 2022-01-07
4902

brendanww:
Handling for standalone extension vs paired extension vs mobile-only walletlink within sendRequest was becoming unwieldy. addEthereumChain, switchEthereumChain, and requestEthereumAccounts all have special logic, so moved special handling for those methods to their own functions so sendRequest remains a more generic function.

Minor change to default null arrays to empty arrays to reduce cognitive overhead.

Return 4902 error so dapps (eg sushi) know to send an `addEthereumChain` request after a `switchEthereumChain` request fails due to the chainId not being supported.

### DIFF
--- a/js/package.json
+++ b/js/package.json
@@ -1,6 +1,6 @@
 {
   "name": "walletlink",
-  "version": "2.4.0",
+  "version": "2.4.1",
   "description": "WalletLink JavaScript SDK",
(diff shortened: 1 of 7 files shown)


khill25 (approved):

01SMD (commented):
.

PR #276, opened 2022-01-15
Delay loading relay

hieronymus777:
WalletLinkRelay opens a websocket immediately in its constructor.  This causes unnecessary network usage since dapps might construct the wallet link web3 provider without using. 

WalletLinkProvider already gets WalletLinkRelay lazily so this is a smallish change to make sure we're not constructing WalletLinkRelay before it's used.

One note, in the case where a walletlink connection has been linked (eip 1102 complete), WalletLinkProvider will load WalletLinkRelay immediately on construction.  That way we can receive any pending responses from mobile clients.

NOTE: still a draft as it needs more testing

### DIFF
--- a/js/package.json
+++ b/js/package.json
@@ -1,6 +1,6 @@
 {
   "name": "walletlink",
-  "version": "2.4.2",
+  "version": "2.4.3",
   "description": "WalletLink JavaScript SDK",
   "keywords": [
     "cipher",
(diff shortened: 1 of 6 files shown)


bangtoven:
@hieronymus777 are you working on this?

xboehmx33 (commented):
a

bangtoven:
closing it since it's been inactive for 180+ days. please reopen if needed.

PR #283, opened 2022-01-21
Fix/unlinked error event

hieronymus777:
We were incorrectly logging this error event for users who didn't sign in to extension through walletlink


brendanww (approved):
Good catch!

Might not be a bad idea if we define "IsStandaloneSigning" as a constant w/i walletlink and reference in the standalone extension code when writing.

PR #302, opened 2022-02-02
Remove backwards compatibility code

brendanww:
`AddEthereumChainResponse` and `SwitchEthereumChainResponse` had handling for two types of response that can come from the clients. New clients start rolling out on Monday February 7. It'll take 2-4 weeks for our next walletlink update to percolate to dapps, which is the amount of time needed for new clients to hit 99% adoption.

### DIFF
--- a/js/src/provider/WalletLinkProvider.ts
+++ b/js/src/provider/WalletLinkProvider.ts
@@ -7,7 +7,6 @@ import BN from "bn.js"
 import { ethErrors } from "eth-rpc-errors"
 
 import { WalletLinkAnalytics } from "../connection/WalletLinkAnalytics"
-import { EthereumChain } from "../EthereumChain"
 import { EVENTS, WalletLinkAnalyticsAbstract } from "../init"
 import { ScopedLocalStorage } from "../lib/ScopedLocalStorage"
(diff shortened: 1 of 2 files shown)


js/src/provider/WalletLinkProvider.ts
```diff
@@ -293,20 +283,10 @@ export class WalletLinkProvider
...
-      const switchResponse = res.result as SwitchResponse
-      if (switchResponse.isApproved && switchResponse.rpcUrl.length > 0) {
-        this._storage.setItem(HAS_CHAIN_BEEN_SWITCHED_KEY, "true")
-        this.updateProviderInfo(switchResponse.rpcUrl, chainId, false)
-      }
-    } else {
-      // this is for legacy clients that return a boolean as result. can deprecate below in February 2022
-      if (res.result) {
-        this._storage.setItem(HAS_CHAIN_BEEN_SWITCHED_KEY, "true")
-        const ethereumChain = EthereumChain.fromChainId(BigInt(chainId))!
```
fan-zhang-sv, line 306: we could remove the import for EthereumChain at line 10

fan-zhang-sv (approved):

PR #315, opened 2022-02-07
Guarantee lowercase addresses are compared

brendanww:
Addresses https://github.com/walletlink/coinbase-wallet-sdk/issues/303

I believe `_addresses` is being mutated by the dapp

https://www.typescriptlang.org/play?ssl=10&ssc=27&pln=1&pc=1#code/DYUwLgBA+ghgJnATiAzi1EC8EDaAiABgA8ZSy8BdAKCoDMBXAOwGMwBLAe0YgHNwBBBMjSoAFAEoIAbyoQ5EZGHqJusIanQoqAXxoB6PRGYc4ICG0Yo2piHBgAHexAC29MDDAY1SDRhCMwRABPKlBIeB8RFCxeAXUoiSoI4U0cAgoYwiJ+HNy8KmNLDlAAOmAOHlFvFNQ0inEgA

```
let _addresses = ["0xaaaaaa"]

function getAddresses() {
    return _addresses
}

// code inside dapp mutates _addresses entry
let addresses = getAddresses()
addresses[0] = "0xAAAAAA"
console.log(_addresses[0])
```

### DIFF
--- a/js/package.json
+++ b/js/package.json
@@ -66,7 +66,7 @@
     "eslint-plugin-prettier": "^4.0.0",
     "eslint-plugin-simple-import-sort": "^7.0.0",
     "jasmine": "3.8.0",
-    "karma": "^6.3.2",
+    "karma": "^6.3.15",
     "karma-browserify": "8.1.0",
     "karma-chrome-launcher": "^3.1.0",
     "karma-jasmine": "^4.0.1",
--- a/js/package.json
+++ b/js/package.json
@@ -1,6 +1,6 @@
 {
   "name": "walletlink",
-  "version": "2.4.6",
+  "version": "2.4.7",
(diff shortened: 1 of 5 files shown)


js/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "walletlink",
-  "version": "2.4.6",
+  "version": "2.4.7",
```
cb-jake, line 3: Do you need to update the `version.ts` for the new version

brendanww: version.ts is auto-generated when `yarn build:prod` is run

cb-jake: ahh thats right, its part of the build step!

hieronymus777 (approved):

PR #336, opened 2022-02-15
feat: add wallet_watchAsset eht request support

pedro-rosario:
Today, one of the major uses of Ethereum wallets is to acquire and track assets. Currently, each wallet either needs to pre-load a list of approved assets, or users need to be stepped through a tedious process of adding an asset for their given wallet.

In the first case, wallets are burdened with both the security of managing this list, as well as the bandwidth of mass polling for known assets on their wallet.

In the second case, the user experience is terrible.

The [wallet_watchAsset](https://eips.ethereum.org/EIPS/eip-747) is a RPC method for allowing users to easily track new assets with a suggestion from sites they are visiting.

### DIFF
--- a/README.md
+++ b/README.md
@@ -151,8 +151,8 @@ Use [EIP-1102](https://eips.ethereum.org/EIPS/eip-1102) to obtain authorization 
 The following code runs in response to a user-initiated action such as clicking a button to ensure the pop up is not blocked by the browser.
 
 ```typescript
-// Use eth_RequestAccounts
-ethereum.request('eth_requestAccounts').then((accounts: string[]) => {
+// Use eth_requestAccounts
+ethereum.send('eth_requestAccounts').then((accounts: string[]) => {
   console.log(`User's address is ${accounts[0]}`)
 
   // Optionally, have the default account set for web3.js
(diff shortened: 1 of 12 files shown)


src/provider/JSONRPC.ts
```diff
@@ -40,7 +41,7 @@ export enum JSONRPCMethod {
   eth_getFilterLogs = "eth_getFilterLogs"
 }
 
-export interface JSONRPCRequest<T = any[]> {
+export interface JSONRPCRequest<T = any> {
```
pedro-rosario, line 44: Changing here in order to avoid a typescript error on the dapps side. On wallet_watchAsset ethereum request, the field params is not an array (i'm not sure why it has changed for this one), so JSONRPCRequest now should expect both arrays and objects.

[image](https://user-images.githubusercontent.com/26755665/154082241-5d72dc25-ce35-48c3-8a9e-60d7d32be126.png)

Ref: [EIP-747](https://eips.ethereum.org/EIPS/eip-747),

TANAWATTHITIWIT:
Hello​

README.md
```diff
@@ -151,8 +151,8 @@ Use [EIP-1102](https://eips.ethereum.org/EIPS/eip-1102) to obtain authorization
 The following code runs in response to a user-initiated action such as clicking a button to ensure the pop up is not blocked by the browser.
 
 ```typescript
-// Use eth_RequestAccounts
-ethereum.request('eth_requestAccounts').then((accounts: string[]) => {
+// Use eth_requestAccounts
+ethereum.send('eth_requestAccounts').then((accounts: string[]) => {
```
brendanww, line 155: @pedro-rosario why is this reverted? request is preferred to send

README.md
```diff
@@ -263,7 +263,79 @@ supported. Here's an example:
...
+  ,
+})
+.then((result: boolean) =>
+  result ? onApproveWatchAsset() : onDenyWatchAsset()
+)
+.catch(err => onError(err.message));
+
+
+// Alternatively, you can use ethereum.send
+ethereum
```
brendanww, line 321: `send` is deprecated, we should probably avoid including it in the docs

brendanww (dismissed):
looks good aside from a few comments

README.md
pedrorosarioo: That was a mistake, only line 154 should be changed, it is a typo fix

brendanww (approved):

brendanww (approved):

README.md
pedro-rosario: That was a mistake, only line 154 should be changed, it is a typo fix

PR #365, opened 2022-02-25
[Fix] Add ethereum chain should work if the user hasn't authed yet

hieronymus777:
First do an eip1102 as part of add ethereum chain if the wallet isn't yet authorized.

The current implementation will call relay.addEthereumChain even if the wallet isn't authed yet.  This only works in standalone (extension) mode.  If coinbase wallet sdk isn't connected to a mobile app yet or the extension isn't signed in yet, this will cause a snack bar to popup that leaves the user in a confusing state.  

Steps to reproduce:
1. With extension signed out and dapp not yet authorized
2. Call add ethereum chain 
Result: snack bar shows telling user to go verify a request on their mobile app
Fixed result: walletlink will ask user to authorize the dapp first.

### DIFF
--- a/src/provider/WalletLinkProvider.ts
+++ b/src/provider/WalletLinkProvider.ts
@@ -269,10 +269,11 @@ export class WalletLinkProvider
     }
 
     const relay = await this.initializeRelay()
+      const isWhitelistedNetwork = !relay.supportsUnauthedAddEthereumChain(chainId.toString())
 
     if (
       !this._isAuthorized() &&
-      !relay.supportsUnauthedAddEthereumChain(chainId.toString())
+      isWhitelistedNetwork
     ) {
       await relay.requestEthereumAccounts().promise
     }
--- a/src/provider/WalletLinkProvider.ts
+++ b/src/provider/WalletLinkProvider.ts
@@ -269,11 +269,11 @@ export class WalletLinkProvider
     }
 
     const relay = await this.initializeRelay()
-      const isWhitelistedNetwork = !relay.supportsUnauthedAddEthereumChain(chainId.toString())
+      const isWhitelistedNetworkOrStandalone = relay.inlineAddEthereumChain(chainId.toString())
 
     if (
       !this._isAuthorized() &&
-      isWhitelistedNetwork
+      !isWhitelistedNetworkOrStandalone 
     ) {
       await relay.requestEthereumAccounts().promise
     }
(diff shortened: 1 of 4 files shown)


cb-kfroemming (approved):

cb-jake (approved):

cb-kfroemming (approved):

cb-kfroemming (approved):

cb-kfroemming (approved):

src/relay/WalletLinkRelayAbstract.ts
```diff
@@ -106,6 +106,12 @@ export abstract class WalletLinkRelayAbstract {
     chainIdCallback: (chainId: string, jsonRpcUrl: string) => void
   ): void
 
+  /**
+   * Whether the provider should first call request ethereum accounts
+   * if handling an add ehereum chain call
```
brendanww, line 111: nit: typo `ehereum` -> `ethereum`

src/provider/WalletLinkProvider.ts
```diff
@@ -269,6 +269,14 @@ export class WalletLinkProvider
     }
 
     const relay = await this.initializeRelay()
+
+    if (
+      !this._isAuthorized() &&
+      !relay.supportsUnauthedAddEthereumChain(chainId.toString())
```
brendanww, line 275: intent would be more clear to me if the code were written as

```
const isWhitelistedNetwork = this.ui.inlineAddEthereumChain(chainId)
if (!this._isAuthorized() && !isWhitelistedNetwork)
```

cb-kfroemming: Updated 👍

src/relay/WalletLinkRelayAbstract.ts
cb-kfroemming: updated 👍

brendanww (approved):

cb-jake (approved):

src/provider/WalletLinkProvider.ts
```diff
@@ -269,6 +269,15 @@ export class WalletLinkProvider
     }
 
     const relay = await this.initializeRelay()
+      const isWhitelistedNetwork = !relay.supportsUnauthedAddEthereumChain(chainId.toString())
```
hieronymus777, line 272: Sorry I think this is more confusing :D .  The reason is that it's not really a boolean for is whitelisted network 
`supportsUnauthedAddEthereumChain` = `true` if `white listed network || standalone extension`

I suggest just doing this:
WalletLinkRelay:
```
inlineAddEthereumChain(chain: string): boolean {
  return this.ui.inlineAddEthereumChain(chain)
}
```

```
const isWhiteListedNetworkOrStandalone = relay.inlineAddEthereumChain(chainId)
```

cb-kfroemming (approved):

PR #369, opened 2022-02-25
Clean up extension.isCipher checks & sample extension

bangtoven:


src/WalletLink.ts
```diff
@@ -185,10 +173,21 @@ export class WalletLink {
...
+    const extension = this.getCBWalletExtension()
+    if (extension) {
+      extension.close()
     } else {
       this._relay?.resetAndReload()
     }
   }
+
+  private getCBWalletExtension(): WalletLinkProvider | undefined {
+    return window.walletLinkExtension
```
bangtoven, line 185: this will include the new variable added for rebranding and resolve whether extension is installed.
```suggestion
    return window.walletLinkExtension || window.coinbaseWallet
```

src/WalletLink.ts
```diff
@@ -185,10 +173,21 @@ export class WalletLink {
...
+  }
+
+  private isCipherProvider(provider: WalletLinkProvider): boolean {
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore
+    return typeof provider.isCipher === "boolean" && provider.isCipher
```
cb-jake, line 191: I don't think `isCipher` is used anymore.

src/WalletLink.ts
```diff
@@ -185,10 +173,21 @@ export class WalletLink {
...
-      window.walletLinkExtension.close()
+    const extension = this.getCBWalletExtension()
+    if (extension) {
+      extension.close()
     } else {
       this._relay?.resetAndReload()
     }
   }
+
+  private getCBWalletExtension(): WalletLinkProvider | undefined {
```
cb-jake, line 184: Would a getter be better than a private method?

```ts
get walletExtenstion () {
    return window.walletLinkExtension
}
```

src/WalletLink.ts
cb-jake, line 191: Is `isCipher` used anymore? I don't see any reference to it in cb wallet.

src/provider/WalletLinkProvider.ts
```diff
@@ -530,9 +530,6 @@ export class WalletLinkProvider
     this._addresses = newAddresses
     this.emit("accountsChanged", this._addresses)
     this._storage.setItem(LOCAL_STORAGE_ADDRESSES_KEY, newAddresses.join(" "))
-    window.dispatchEvent(
```
cb-jake, line 533: Is this not needed anymore? How does this affect the wallet products that use this provider?

cb-jake (commented):
@bangtoven just a few questions.

src/WalletLink.ts
bangtoven: nice

src/WalletLink.ts
bangtoven: correct. it's not in our current code base. i think this is for backward compatibility to support old clients who haven't updated yet.

src/provider/WalletLinkProvider.ts
bangtoven: this was used by the sample project only, under `chrome` folder.

bangtoven:
@cb-jake can you take a look again? 🙏

cb-jake (approved):
lgtm. Might be good to get @hieronymus777's eyes on it as well.

bangtoven:
@hieronymus777 merging it for now. lmk if you have any concerns.

PR #381, opened 2022-03-05
Fix: Send dapps current network chainId through walletlink on wallet_watchAsset event

pedro-rosario:
For this first version of `watchAsset` feature, we're assuming that once users authorize wallet_watchAsset request, the token should be added in the current network. That said, we've decided to follow the approach of sending its current network chainId on the event params (cc @brendanww). 

It may be used by both extension and mobile apps to add token in the correct network

### DIFF
--- a/src/provider/CoinbaseWalletProvider.ts
+++ b/src/provider/CoinbaseWalletProvider.ts
@@ -235,7 +235,8 @@ export class CoinbaseWalletProvider
     address: string,
     symbol?: string,
     decimals?: number,
-    image?: string
+    image?: string,
+    chainId?: number
   ): Promise<boolean> {
     const relay = await this.initializeRelay()
     const result = await relay.watchAsset(
@@ -243,7 +244,8 @@ export class CoinbaseWalletProvider
       address,
       symbol,
       decimals,
-      image
+      image,
+      chainId?.toString()
     ).promise
 
     return !!result.result
@@ -1124,6 +1126,7 @@ export class CoinbaseWalletProvider
       })
     }
 
+    const chainId = parseInt(this.chainId, 16)
     const { address, symbol, image, decimals } = request.options
 
     const res = await this.watchAsset(
@@ -1131,7 +1134,8 @@ export class CoinbaseWalletProvider
       address,
       symbol,
       decimals,
-      image
+      image,
+      chainId
     )
 
     return { jsonrpc: "2.0", id: 0, result: res }
--- a/src/provider/CoinbaseWalletProvider.ts
+++ b/src/provider/CoinbaseWalletProvider.ts
@@ -1126,7 +1126,7 @@ export class CoinbaseWalletProvider
       })
     }
 
-    const chainId = parseInt(this.chainId, 16)
+    const chainId = this.getChainId();
     const { address, symbol, image, decimals } = request.options
 
     const res = await this.watchAsset(
--- a/src/provider/WalletUI.ts
+++ b/src/provider/WalletUI.ts
@@ -58,6 +58,7 @@ export interface WalletUI {
     symbol?: string
     decimals?: number
     image?: string
+    chainId?: string
   }): void
 
   switchEthereumChain(options: {
--- a/src/relay/WalletSDKRelay.ts
+++ b/src/relay/WalletSDKRelay.ts
@@ -78,9 +78,7 @@ export interface WalletSDKRelayOptions {
   darkMode: boolean
   storage: ScopedLocalStorage
   relayEventManager: WalletSDKRelayEventManager
-  uiConstructor: (
-    options: Readonly<WalletUIOptions>
-  ) => WalletUI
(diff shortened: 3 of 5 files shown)


src/version.ts
```diff
@@ -1 +1 @@
-export const LIB_VERSION = "3.0.1";
+export const LIB_VERSION = "3.0.2"
```
ljharb, line 1: typically version numbers aren't updated in PRs, only when releasing a new version

src/provider/CoinbaseWalletProvider.ts
```diff
@@ -235,15 +235,17 @@ export class CoinbaseWalletProvider
...
   ): Promise<boolean> {
     const relay = await this.initializeRelay()
     const result = await relay.watchAsset(
       type,
       address,
       symbol,
       decimals,
-      image
+      image,
+      chainId?.toString()
```
ljharb, line 248:
```suggestion
      chainid ?? String(chainid)
```
since `.toString()` doesn't guarantee a string return value, but `String()` does

pedro-rosario: Not sure if we should, we don't want to guarantee a string return value. In cases where it doesn't find a chainId, we just shouldn't pass any string, that's why it is an optional param. To pass "null" or "undefined" (as strings) wouldn't help in this case since it wouldn't refer to any network.

mikehunter1990: How

cb-jake (approved):
lgtm

bangtoven (dismissed):

src/provider/CoinbaseWalletProvider.ts
```diff
@@ -1124,14 +1126,16 @@ export class CoinbaseWalletProvider
       })
     }
 
+    const chainId = parseInt(this.chainId, 16)
```
bangtoven, line 1129: seems we could use `getChainId` function directly without parsing it to an int @pedro-rosario

bangtoven: `this.chainId` is a formatted version of `getChainId`
image

src/provider/CoinbaseWalletProvider.ts
ljharb: The issue is that any object can have a `toString` method that doesn't return a string.

With my suggestion, it will return `null` or `undefined` if those are what's passed, otherwise, it will *guarantee* a string return.

erin-at-work (approved):

cb-heimdall:
Review Error for erin-at-work @ 2022-03-08 13:52:01 UTC

src/provider/CoinbaseWalletProvider.ts
jeongmincho: not sure what we're looking for on the frontend, but if `undefined` is passed in `chainid ?? String(chainid)` would return the string `'undefined'`. not sure if i'm following the point about any object being passed in here - wouldn't it just be a number?

jeongmincho (approved):

src/provider/CoinbaseWalletProvider.ts
ljharb: oh right, sorry - the suggestion should be:
```suggestion
      chainId == null ? chainId : String(chainId)
```

This is JS; any kind of value could be passed in.

cb-jake (approved):
+1 once @ljharb comment is resolved

src/provider/CoinbaseWalletProvider.ts
pedro-rosario: > not sure what we're looking for on the frontend, but if `undefined` is passed in `chainid ?? String(chainid)` would return the string `'undefined'`. not sure if i'm following the point about any object being passed in here - wouldn't it just be a number?

@jeongmincho Actually it wouldn't happen because if chainId is a falsy value, it would return undefined/null and never process `String(chainId)`




> The issue is that any object can have a `toString` method that doesn't return a string.
> 
> With my suggestion, it will return `null` or `undefined` if those are what's passed, otherwise, it will _guarantee_ a string return.

@ljharb, Isn't it already guaranteed here https://github.com/coinbase/coinbase-wallet-sdk/blob/9c58ef59625d0dba7d5c151ea984815fe8c61259/src/provider/CoinbaseWalletProvider.ts#L854-L858?

bangtoven (approved):

src/provider/CoinbaseWalletProvider.ts
ljharb: no - `chainId` is a prop passed in by a user. a prop can be literally any value; TS doesn't exist at runtime.

What you linked guarantees that you only *pass* an integer number - but users can pass anything they want.

jeongmincho: @pedro-rosario `??` returns the right-hand side if left-hand side is undefined/null

pedro-rosario: Oh sry, you're right

PR #410, opened 2022-03-16
Disable address switch

hieronymus777:
Add option to disable switching selected address.  Currently the extension doesn't support this gracefully leading to timing issues if you connect a dapp to your mobile phone via the extension, and the mobile has a different address selected than the one the extension connected to originally.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@coinbase/wallet-sdk",
-  "version": "3.0.5",
+  "version": "3.0.4",
   "description": "Coinbase Wallet JavaScript SDK",
   "keywords": [
     "cipher",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@coinbase/wallet-sdk",
-  "version": "3.0.4",
+  "version": "3.0.5",
   "description": "Coinbase Wallet JavaScript SDK",
   "keywords": [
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@coinbase/wallet-sdk",
-  "version": "3.0.4",
+  "version": "3.0.5",
```
ljharb, line 3: versions aren't bumped in PRs
```suggestion
  "version": "3.0.4",
```

src/provider/CoinbaseWalletProvider.ts
```diff
@@ -81,6 +82,8 @@ export class CoinbaseWalletProvider
 
   private hasMadeFirstChainChangedEmission = false
 
+  private _supportsAddressSwitching?: boolean
```
ljharb, line 85: the leading underscore seems redundant with TS's `private`

ljharb (commented):
tests?

src/provider/CoinbaseWalletProvider.ts
hieronymus777: 7f6f2a59fd1ea9fa93842c9122a4da01609ff921

jeongmincho (dismissed):

bangtoven (dismissed):

PR #412, opened 2022-03-17
Install CoinbaseWallet dialog user response

erin-at-work:
- Updates text when user clicks on 'Install' -> Prompts user to connect again after clicking to refresh the page.
- Also opens mobile app download page in a new window

image

### DIFF
--- a/src/components/TryExtensionLinkDialog.tsx
+++ b/src/components/TryExtensionLinkDialog.tsx
@@ -2,6 +2,7 @@ import clsx from "clsx"
 import { FunctionComponent, h } from "preact"
 import { useEffect, useState } from "preact/hooks"
 
+import { LIB_VERSION } from "../version"
 import globeIcon from "./icons/globe-icon-svg"
 import linkIcon from "./icons/link-icon-svg"
 import lockIcon from "./icons/lock-icon-svg"
@@ -9,7 +10,6 @@ import walletLogo from "./icons/QRLogo"
 import { QRCode } from "./QRCode"
 import { Spinner } from "./Spinner"
 import css from "./TryExtensionLinkDialog-css"
-import {LIB_VERSION} from "../version";
 
 export const TryExtensionLinkDialog: FunctionComponent<{
   darkMode: boolean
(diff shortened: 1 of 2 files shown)


bangtoven (dismissed):

src/components/TryExtensionLinkDialog.tsx
```diff
@@ -100,11 +100,30 @@ export const TryExtensionLinkDialog: FunctionComponent<{
 const TryExtensionBox: FunctionComponent<{
   onInstallClick: () => void
 }> = props => {
+  const [isClicked, setIsClicked] = useState(false)
+
+  const clickHandler = () => {
+    if (isClicked) {
+      window.location.reload()
+    } else {
+      props.onInstallClick()
```
ljharb, line 109:
```suggestion
}> = ({ onInstallClick }) => {
  const [isClicked, setIsClicked] = useState(false)

  const clickHandler = () => {
    if (isClicked) {
      window.location.reload()
    } else {
      onInstallClick();
```
otherwise, the `props` object is passed as the receiver to the `onInstallClick` prop, which is not something that should be exposed.

src/components/TryExtensionLinkDialog.tsx
```diff
@@ -100,11 +100,30 @@ export const TryExtensionLinkDialog: FunctionComponent<{
...
       <div class="-cbwsdk-extension-dialog-box-top-install-region">
         <h2>Try the Coinbase Wallet extension</h2>
-        <button onClick={props.onInstallClick}>Install</button>
+        {isClicked && (
+          <div class="-cbwsdk-extension-dialog-box-top-subtext">
+            After installing Coinbase Wallet, refresh the page and connect
+            again.
+          </div>
+        )}
+        <button onClick={clickHandler}>
```
ljharb, line 124:
```suggestion
        <button type="button" onClick={clickHandler}>
```
all button elements should always have an explicit type, because the default type is `submit`

src/components/TryExtensionLinkDialog.tsx
```diff
@@ -100,11 +100,30 @@ export const TryExtensionLinkDialog: FunctionComponent<{
...
+
   return (
     <div class="-cbwsdk-extension-dialog-box-top">
       <div class="-cbwsdk-extension-dialog-box-top-install-region">
         <h2>Try the Coinbase Wallet extension</h2>
-        <button onClick={props.onInstallClick}>Install</button>
+        {isClicked && (
+          <div class="-cbwsdk-extension-dialog-box-top-subtext">
+            After installing Coinbase Wallet, refresh the page and connect
+            again.
```
ljharb, line 121: nit:
```suggestion
            After installing Coinbase Wallet, refresh the page and connect again.
```

erin-at-work: image
 
I think this is due to our prettier setting, `printWidth` is set at 80

jeongmincho (approved):

PR #413, opened 2022-03-17
Formatting: eslint + prettier

erin-at-work:
-  Formats current files


.husky/pre-commit
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
```
ljharb, line 1: i would actually strongly discourage husky; git hooks included in the repo are a security concern, which is why you can't do this directly in git, and in the past i've seen husky inside node_modules install git hooks in a corporate internal repo.

package.json
```diff
@@ -66,11 +67,13 @@
     "eslint-plugin-preact": "^0.1.0",
     "eslint-plugin-prettier": "^4.0.0",
     "eslint-plugin-simple-import-sort": "^7.0.0",
+    "husky": "7.0.4",
```
ljharb, line 70:
```suggestion
    "husky": "^7.0.4",
```
altho i'd suggest removing it, all deps should always use `^`

ljharb (commented):
linting should be done in CI, and isn't important to do anywhere else - and anything not enforced in CI isn't enforced at all, since local git hooks can always be bypassed.

.husky/pre-commit
erin-at-work: Noted!

> linting should be done in CI, and isn't important to do anywhere else - and anything not enforced in CI isn't enforced at all, since local git hooks can always be bypassed.

Yes, ideally we'd have this setup in our CI which we still need to setup. I can remove husky in favor of `.githooks/pre-push`

ljharb: To be clear, I'm suggesting avoiding any git hooks in the repo whatsoever, as they are always a security risk - it's just that husky is particularly dangerous in my experience.

.eslintrc.js
```diff
@@ -1,224 +1,208 @@
...
-    "env": {
-        "browser": true,
-        "es6": true,
-        "node": true
-    },
-    "extends": [
-        "plugin:@typescript-eslint/recommended",
-        "plugin:@typescript-eslint/recommended-requiring-type-checking",
-        "prettier"
+  env: {
```
ljharb, line 2:
```suggestion
  root: true,
  env: {
```
so that any eslint configs in a parent directory aren't factored in

.husky/pre-commit
erin-at-work: @ljharb Got it. Thanks for the context. I'm going to limit this PR to formatting & linting of current files

ljharb: For setting up CI, i'd be happy to pair with you on setting up github actions, which is already enabled for this org.

erin-at-work: that'd be great! will reach out to setup some time

karma.conf.js
```diff
@@ -1,25 +1,21 @@
...
-    plugins: [
-      'karma-jasmine',
-      'karma-chrome-launcher',
-      'karma-browserify'
-    ],
+    plugins: ["karma-jasmine", "karma-chrome-launcher", "karma-browserify"],
 
     // base path that will be used to resolve all patterns (eg. files, exclude)
-    basePath: '',
+    basePath: "",
```
ljharb, line 6: shouldn't we be using single quotes for JS strings?

src/CoinbaseWalletSDK.ts
```diff
@@ -10,12 +10,9 @@ import { WalletSDKRelay } from "./relay/WalletSDKRelay"
 import { WalletSDKRelayEventManager } from "./relay/WalletSDKRelayEventManager"
 import { getFavicon } from "./util"
 
-const LINK_API_URL =
-    process.env.LINK_API_URL! || "https://www.walletlink.org"
+const LINK_API_URL = process.env.LINK_API_URL! || "https://www.walletlink.org"
```
ljharb, line 13: unrelated to this PR, why is the `!` necessary? the `||` should tell TS that the type of the var is `string`

src/version.ts
```diff
@@ -1 +1 @@
-export const LIB_VERSION = "3.0.4";
+export const LIB_VERSION = "3.0.4"
```
ljharb, line 1: there's a ton of reasons why semicolon usage should be enforced; can that be enabled in the prettier config? fine to do in a followup too

bangtoven (approved):

cb-heimdall:
Review Error for bangtoven @ 2022-03-18 17:38:55 UTC

src/CoinbaseWalletSDK.ts
erin-at-work: Good catch. I haven't authored 99% of the current codebase so it's likely much of the code were not caught in code reviews or even lint rules.

src/version.ts
erin-at-work: I noticed that too. Will enable in a separate PR

bangtoven (approved):

Issue #343, opened 2022-02-17
Incompatibility issue with Web3Modal in Angular


karthedew:
This is an interesting issue because it works in a React application in scaffold-eth, but not in Angular. The issue originates in declaring `ethereum` globally as an extension of the Window object. In both WalletLink and Web3Modal, ethereum is extended on the window, but the interfaces declare *ethereum* differently. When compiling an Angular application, it catches this discrepancy and reports it as an error.

I'm not sure what the right path is. Hoping for some resolution on this without having to change the source code of this package in my project.

WalletLink's global window:
https://github.com/coinbase/coinbase-wallet-sdk/blob/master/src/index.ts
```
declare global {
  interface Window {
    WalletLink: typeof WalletLink
    WalletLinkProvider: typeof WalletLinkProvider
    ethereum?: WalletLinkProvider
    walletLinkExtension?: WalletLinkProvider
  }
}
```

Web3Modal global window:
https://github.com/Web3Modal/web3modal/blob/master/src/components/Modal.tsx
```
declare global {
  // tslint:disable-next-line
  interface Window {
    ethereum: any;
    BinanceChain: any;
    web3: any;
    celo: any;
    updateWeb3Modal: any;
  }
}
```

ewerx:
Thanks for the report, we will investigate a solution.

christophercr:
Same issue here.

AntonioCardenas:
Same issue here.
And I'm using this options in ``tsconfig.json`` to allow the crypto usage and avoid common errors related to webpack 5: 
``` tsconfig.json
  "compilerOptions": {
    "paths":{
      "crypto": ["./node_modules/crypto-browserify"],
      "stream": ["./node_modules/stream-browserify"],
      "assert": ["./node_modules/assert"],
      "http": ["./node_modules/stream-http"],
      "https": ["./node_modules/https-browserify"],
      "url": ["./node_modules/url"],
      "os": ["./node_modules/os-browserify"],
    },
```
**The error output is** 

```
 TS2717: Subsequent property declarations must have the same type.  Property 'ethereum' must be of type 'any', but here has type 'CoinbaseWalletProvider'.

10         ethereum?: CoinbaseWalletProvider;
           ~~~~~~~~

  node_modules/web3modal/dist/components/Modal.d.ts:6:9
    6         ethereum: any;
              ~~~~~~~~
    'ethereum' was also declared here.


✖ Failed to compile.
[...]

PR #432, opened 2022-03-25
Update window.ethereum type to unknown

erin-at-work:
### *Summary*

- Prevent conflicting wallet providers for `window.ethereum`

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -12,11 +12,7 @@ declare global {
   interface Window {
     CoinbaseWalletSDK: typeof CoinbaseWalletSDK;
     CoinbaseWalletProvider: typeof CoinbaseWalletProvider;
-    /**
-     * Workaround possible compatibility issues in Angular
-     * Set as CoinbaseWalletProvider
-     */
-    ethereum?: any;
+    ethereum?: unknown;
     coinbaseWalletExtension?: CoinbaseWalletProvider;
 
     // deprecated
--- a/src/index.ts
+++ b/src/index.ts
@@ -22,7 +22,13 @@ declare global {
      * @deprecated Legacy API
      */
     WalletLink: typeof CoinbaseWalletSDK;
+    /**
+     * @deprecated Legacy API
+     */
     WalletLinkProvider: typeof CoinbaseWalletProvider;
+    /**
+     * @deprecated Legacy API
+     */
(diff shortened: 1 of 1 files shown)


src/index.ts
```diff
@@ -12,7 +12,11 @@ declare global {
   interface Window {
     CoinbaseWalletSDK: typeof CoinbaseWalletSDK;
     CoinbaseWalletProvider: typeof CoinbaseWalletProvider;
-    ethereum?: CoinbaseWalletProvider;
+    /**
+     * Workaround possible compatibility issues in Angular
+     * Set as CoinbaseWalletProvider
+     */
+    ethereum?: any;
```
erin-at-work, line 19: This should probably be `unknown`?

src/index.ts
```diff
@@ -12,11 +12,7 @@ declare global {
   interface Window {
     CoinbaseWalletSDK: typeof CoinbaseWalletSDK;
     CoinbaseWalletProvider: typeof CoinbaseWalletProvider;
-    /**
-     * Workaround possible compatibility issues in Angular
-     * Set as CoinbaseWalletProvider
-     */
-    ethereum?: any;
+    ethereum?: unknown;
```
ljharb, line 15: i wonder if we could instead declare `unknown | PreciseType` so that type hinting is more useful.

erin-at-work: I had first attempted that, but it doesn't seem to get picked up. 
> // In a union an unknown absorbs everything

from https://github.com/microsoft/TypeScript/pull/24439

image

and `coinbaseWalletExtension`:
image

I'll add back a comment

ljharb: ah ok, oof then.

erin-at-work: @ljharb Thoughts on addressing this issue? https://github.com/coinbase/coinbase-wallet-sdk/issues/343

> All declarations of 'ethereum' must have identical modifiers

In order to fix the above error, the type would need to be updated to:

```ts
interface Window {
  ethereum: any | undefined;
}
```

But I hesitate to make changes based on a conflicting library's type defs

bangtoven (approved):

cb-heimdall:
Review Error for bangtoven @ 2022-03-25 19:27:40 UTC

bangtoven (approved):

PR #444, opened 2022-04-05
Added Get QR URL Method 

jeongmincho:
### *Summary*

- Expose QR URI as a public readonly field in `CoinbaseWalletSDK`

- Locally tested with dummy dapp


src/CoinbaseWalletSDK.ts
```diff
@@ -92,6 +93,9 @@ export class CoinbaseWalletSDK {
     });
     this.setAppInfo(options.appName, options.appLogoUrl);
     this._relay.attachUI();
+
+    const serverUrl = window.encodeURIComponent(linkAPIUrl);
+    this._qrUrl = `${linkAPIUrl}/#/link?id=${this._relay.session.id}&secret=${this._relay.session.secret}&server=${serverUrl}&v=1`;
```
bangtoven, line 98: let's create a dedicated class or function or module for the sake of DRY.
so i mean, refactoring `ScanQRCode` and `ScanQRBox` accordingly as well.

ljharb: also it'd be great to avoid manually building a query string, and instead to use an actual parser/generator - `URL`, or https://npmjs.com/qs stringify

jeongmincho: I caught up with the rainbow eng yesterday and he showed me how they're currently pulling the URI for WalletConnect and how they are fetching CBWSDK (via wagmi's connector). Given that they're using our web3 provider directly from wagmi, instead of importing our SDK library, I'm thinking we can follow the same pattern as WalletConnect and put this QR code uri as a public field on the `CoinbaseWalletProvider`. What do you guys think? @bangtoven @ljharb 

Adding screenshots because it's a private repo until release:

How they fetch WalletConnect URI:
Screen Shot 2022-04-08 at 8 40 31 AM

What getProvider on wagmi does:
Screen Shot 2022-04-08 at 8 41 26 AM

How they use CBWSDK:
Screen Shot 2022-04-08 at 8 44 31 AM

src/CoinbaseWalletSDK.ts
```diff
@@ -92,6 +93,9 @@ export class CoinbaseWalletSDK {
     });
     this.setAppInfo(options.appName, options.appLogoUrl);
     this._relay.attachUI();
+
+    const serverUrl = window.encodeURIComponent(linkAPIUrl);
+    this._qrUri = `${linkAPIUrl}/#/link?id=${this._relay.session.id}&secret=${this._relay.session.secret}&server=${serverUrl}&v=1`;
```
jeongmincho, line 98: i'll update this with `qs`

src/CoinbaseWalletSDK.ts
ljharb: that depends on how walletlink.org is implemented. Using the `/#/` convention is super archaic, and pushState should be used instead.

What I'd expect is that the `?` part is constructed with `qs` or `URLSearchParams`, and then appended to `https://www.walletlink.org/#/link`.

jeongmincho: ok. i've tried something like this:

Screen Shot 2022-04-08 at 12 25 14 PM

but the returning href looks like this: `https://www.walletlink.org/?id=...&secret=...&server=...&v=1#/link` where the `/#/link` is appended at the tail end. 

i'm seeing that these optional fragments are normally placed at the end of the URL, so now i'm thinking the original string is incorrect. thoughts on how to resolve this?
Screen Shot 2022-04-08 at 12 26 47 PM

ljharb: right - because it's behind a fragment, you can't use URLSearchParams on that link. You'd need to do some manual stuff like `new URL('.')`, and then you'd end up with `.?blah`, and then you'd need to slice out the `?blah`, and then append that manually to the actual URL string.

jeongmincho: sounds like the resulting URL construction might end up more difficult to read than the existing interpolated string - i might keep it the way it is for now if that's alright

ljharb: perhaps, but it'd be less bug-prone.

my primary suggestion is this:
```js
this._qrUrl = `${linkAPIUrl}/#/link${qs.stringify({
  id: this._relay.session.id,
  secret :this._relay.session.secret,
  server: serverUrl,
  v: 1,
})}`;
```
which i personally find far more readable.

bangtoven: their current (hacky) method to get url shouldn't affect our design. 
seems wagmi is still using walletlink, so we have to make updates on them anyway. 
let's try to go with a better solution wrt separation of concerns. 
it appears to me that, walletlink.org connection related stuff doesn't belong to provider logic.

ljharb: I agree - but I'm also saying that *always*, building a URL query string should use something like qs or USP :-)

Hlllllllii:
🇯🇵🇧🇷🇰🇷👍 $

src/CoinbaseWalletSDK.ts
jeongmincho: @ljharb i tried using the qs library with the solution you suggested. however, when i stringify the following url (`https%3A%2F%2Fwww.walletlink.org`) it turns into this new url (`https%253A%252F%252Fwww.walletlink.org`) adding a few hex characters (namely, 25) that weren't present in the original string interpolation - might you know how to prevent this? it's breaking the QR linking when scanned with mobile

ljharb: Sorry if i wasn't clear. I don't mean the entire URL, i mean the query string part - just the part that starts with `?`. The rest would be manually prepended to the result.

jeongmincho: No worries, yeah i was just handling the query string part but there is a URL that goes into one of the search params (i.e. the `serverUrl` param which is this string `https%3A%2F%2Fwww.walletlink.org`) wondered if there was a way to work around it without manually writing just the last part out
```
this._qrUrl = `${linkAPIUrl}/#/link${qs.stringify({
  id: this._relay.session.id,
  secret :this._relay.session.secret,
  server: serverUrl,
  v: 1,
})}`;
```

ljharb: not that i know of, unfortunately :-/

src/util.ts
```diff
@@ -221,3 +221,16 @@ export function getFavicon(): string | null {
...
+
+export function createQrUrl(
+  sessionId: string,
+  sessionSecret: string,
+  serverUrl: string,
+  isParentConnection: boolean
+): string {
+  const encodedServerUrl = window.encodeURIComponent(serverUrl);
+  const sessionIdKey = isParentConnection ? "parent-id" : "id";
+  const qrUrl = `${serverUrl}/#/link?${sessionIdKey}=${sessionId}&secret=${sessionSecret}&server=${encodedServerUrl}&v=1`;
```
ljharb, line 233: i'd still prefer to see qs used here, but using this abstraction is *way* clearer than before :-)

ljharb (approved):

src/provider/CoinbaseWalletProvider.ts
```diff
@@ -65,6 +66,7 @@ export class CoinbaseWalletProvider
 {
   // So dapps can easily identify Coinbase Wallet for enabling features like 3085 network switcher menus
   public readonly isCoinbaseWallet: boolean;
+  public readonly qrUrl: string | null = null;
```
bangtoven, line 69: walletlink qr doesn't seem to belong to wallet provider.

bangtoven (commented):
looks good.
- qr url would better be exposed from CoinbaseWalletSDK
- do we need package-lock.json file?

ljharb:
imo packages shouldn't have lockfiles; this package however has yarn.lock.

It would be nice to get rid of it and use package-lock.json instead (ie, use npm instead of yarn) but that's a separate thing.

src/CoinbaseWalletSDK.ts
```diff
@@ -167,6 +175,13 @@ export class CoinbaseWalletSDK {
     }
   }
 
+  /**
+   * Return QR URL for mobile wallet connection, will return null if extension is installed
+   */
+  public getQrUrl(): string | null {
+    return this._qrUrl;
```
erin-at-work, line 182: Will this ever return `null`? or just an empty string? .. checking against `createQrUrl`

jeongmincho: so it's default initiated to null, and when we detect the wallet extension (lines 79-81), it'll not get set and remain null because we won't have created a relay session.
Screen Shot 2022-04-12 at 12 06 21 PM

src/util.ts
bangtoven: +1. @jeongmincho let's address this.

jeongmincho: oh man, you are correct! 🔥

bangtoven (approved):

cb-heimdall:
Review Error for bangtoven @ 2022-04-12 18:01:58 UTC

src/util.ts
```diff
@@ -221,3 +222,23 @@ export function getFavicon(): string | null {
...
+  serverUrl: string,
+  isParentConnection: boolean
+): string {
+  const sessionIdKey = isParentConnection ? "parent-id" : "id";
+
+  const query: string = stringify({
+    [sessionIdKey]: sessionId,
+    secret: sessionSecret,
+    server: serverUrl,
+    v: "1"
```
bangtoven, line 238: how about?
```suggestion
  isParentConnection: boolean,
  version = "1"
): string {
  const sessionIdKey = isParentConnection ? "parent-id" : "id";

  const query: string = stringify({
    [sessionIdKey]: sessionId,
    secret: sessionSecret,
    server: serverUrl,
    v: version
```

bangtoven (approved):

src/util.ts
ljharb: you could even inline it `[isParentConnection ? 'parent-id' : 'id']: sessionId,`

jeongmincho: i think keeping the ternary separate will be a little more readable:)

PR #448, opened 2022-04-06
[fix] Github action for PR creation & tagging

erin-at-work:
### *Summary*

- Update PR title
- Add commit message with tag version
- Fix conditional for tagging

### DIFF
--- a/.github/workflows/versioning.yml
+++ b/.github/workflows/versioning.yml
@@ -45,7 +45,7 @@ jobs:
       - name: Commit and branch updated version
         uses: peter-evans/create-pull-request@v4
         with:
-          title: "[Version update]"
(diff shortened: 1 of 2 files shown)


.github/workflows/versioning.yml
```diff
@@ -45,8 +45,9 @@ jobs:
       - name: Commit and branch updated version
         uses: peter-evans/create-pull-request@v4
         with:
-          title: "v${{ env.PACKAGE_VERSION }}"
+          title: "[Version update]"
```
ljharb, line 48: i think it's actually nice to have the version in the title :-) i just meant it shouldn't be the source of truth
```suggestion
          title: "v${{ env.PACKAGE_VERSION }}"
```

ljharb (approved):

PR #469, opened 2022-04-14
fix: make qr code url available in headless mode

RyanTimesTen:
### *Summary*
Hey y'all, thanks for working on this project 👋 

We'd like to make the QR code URL available when running the SDK in headless mode. Currently there is an [early return in the `CoinbaseWalletSDK` constructor](https://github.com/coinbase/coinbase-wallet-sdk/blob/master/src/CoinbaseWalletSDK.ts#L98) that results in the QR code URL not being set when running in headless mode.

Since the inputs to `createQrUrl` are specific to the relay, it makes sense for the relay to expose it as a method. Users of `CoinbaseWalletSDK` will now be able to call `getQrUrl` and get back a URL whether or not they are running in headless mode!

Local test app

### DIFF
--- a/src/CoinbaseWalletSDK.ts
+++ b/src/CoinbaseWalletSDK.ts
@@ -8,7 +8,7 @@ import { WalletSDKUI } from "./provider/WalletSDKUI";
 import { WalletUI, WalletUIOptions } from "./provider/WalletUI";
 import { WalletSDKRelay } from "./relay/WalletSDKRelay";
 import { WalletSDKRelayEventManager } from "./relay/WalletSDKRelayEventManager";
-import { createQrUrl, getFavicon } from "./util";
+import { getFavicon } from "./util";
 
 const LINK_API_URL = process.env.LINK_API_URL || "https://www.walletlink.org";
 const SDK_VERSION =
@@ -43,7 +43,6 @@ export class CoinbaseWalletSDK {
   private _appLogoUrl: string | null = null;
   private _relay: WalletSDKRelay | null = null;
   private _relayEventManager: WalletSDKRelayEventManager | null = null;
-  private _qrUrl: string | null = null;
   private _storage: ScopedLocalStorage;
   private _overrideIsMetaMask: boolean;
   private _overrideIsCoinbaseWallet: boolean;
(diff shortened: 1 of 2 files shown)


src/CoinbaseWalletSDK.ts
```diff
@@ -184,7 +176,7 @@ export class CoinbaseWalletSDK {
    * Return QR URL for mobile wallet connection, will return null if extension is installed
    */
   public getQrUrl(): string | null {
-    return this._qrUrl;
+    return this._relay?.getQRCodeUrl() ?? null;
```
ljharb, line 179: when would `this._relay` not be set? when would `getQRCodeUrl()` return undefined (and why would we need to coerce undefined to null)?

RyanTimesTen: `this._relay` would not be set because of [this early return in the constructor](https://github.com/coinbase/coinbase-wallet-sdk/blob/master/src/CoinbaseWalletSDK.ts#L81-L83)

RyanTimesTen: since `this._relay` could be `undefined` (because of the early return in the constructor), the expression would resolve to `undefined`, but `getQrUrl` returns `string | null`. So I opted to coerce to `null`. We probably wouldn't want to change the return type of `getQrUrl` at this point

ljharb: ah, i see what you mean, thanks.

RyanTimesTen: actually, `this._relay` could not be `undefined`, it could only be `null` or `WalletSDKRelay`. i believe the optional chaining operator is what causes the expression to resolve to `undefined` if it encounters `null`

RyanTimesTen: confirmed:
image

jeongmincho (approved):
good catch, thanks for the fix!

RyanTimesTen:
@jeongmincho what is the process for getting this fix released?

jeongmincho:
hey @RyanTimesTen just seeing this. 3.0.8 should be published now!

PR #476, opened 2022-04-19
[chore] Setup unit testing

erin-at-work:
### *Summary*

- Init unit testing lib

`yarn test:unit `

image


package.json
```diff
@@ -55,13 +56,17 @@
     "stream-browserify": "^3.0.0"
   },
   "devDependencies": {
-    "@babel/core": "^7.0.0",
+    "@babel/core": "^7.17.9",
```
cb-jake, line 59: can we use `ts-jest` vs introducing babel?

cb-jake (approved):
lgtm. Might be less overhead to use `ts-jest` vs `babel` but I'm not opposed to using `babel`

package.json
ljharb: why? babel is a far superior transpiler to tsc, and should always be used exclusively to transpile, including for TS.

cb-jake: If thats the case, should we upgrade the build to use babel vs tsc?

ljharb: 1000% yes

erin-at-work: ty both! will make that update 👌

PR #479, opened 2022-04-20
[chore] Adds unit test to github action

erin-at-work:
### *Summary*

- Adds unit test to CI

See https://github.com/coinbase/coinbase-wallet-sdk/runs/6098167053?check_suite_focus=true

image


.github/workflows/main.yml
```diff
@@ -32,3 +32,6 @@ jobs:
 
       - name: Check Types
         run: yarn lint:types
+
+      - name: Run Unit Tests
+        run: yarn test:unit
```
ljharb, line 37: probably we want to run this in a matrix, and not in the same job as linting/typechecking

erin-at-work: Matrix seemed like a overkill for PRs so moved for separate check on master btw.

ljharb: I strongly disagree it’s overkill; it’s very common for a PR to work on latest node and break on an older one, which is a semver-major change.

I have 300+ packages, and all of them run 200+ matrix jobs on every push and PR, because i test every supported minor - the minimum imo is testing every supported major.

.github/workflows/ci.yml
```diff
@@ -38,18 +38,13 @@ jobs:
...
-
     steps:
     - name: Checkout
       uses: actions/checkout@v3
 
-    - name: Use Node.js ${{ matrix.node-version }}
+    - name: Use Node.js
       uses: actions/setup-node@v3
       with:
-        node-version: ${{ matrix.node-version }}
```
ljharb: this seems like a pretty important thing to have?

.github/workflows/ci.yml
```diff
@@ -1,20 +1,21 @@
 name: CI
 
 on:
-  pull_request:
-    types: [review_requested]
+  push:
+    branches-ignore:
+      - master
```
ljharb, line 6: I’m confused; why not run everything on everything? Linting matters in master just as much as on PRs, and still could fail after two passing PRs are merged.

.github/workflows/main.yml
```diff
@@ -1,20 +1,19 @@
 name: CI
 
 on:
-  pull_request:
-    types: [review_requested]
+  push
```
ljharb, line 4:
```suggestion
  [push, pull_request]
```
so it runs on both

.github/workflows/main.yml
```diff
@@ -32,3 +31,26 @@ jobs:
...
+          node-version: [12.x, 14.x, 16.x]
+
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v3
+
+    - name: Use Node.js ${{ matrix.node-version }}
+      uses: actions/setup-node@v3
+      with:
+        cache: 'yarn'
```
ljharb, line 50: i'm not sure how this can implicitly use the matrix's node version - i think you need this:
```suggestion
    - name: Use Node.js ${{ matrix.node-version }}
      uses: actions/setup-node@v3
      with:
        node-version: ${{ matrix.node-version }}
        cache: 'yarn'
```

erin-at-work: Ah yup, good catch.. missed it in a cleanup

ljharb (approved):
LGTM!

brianchitester (approved):

PR #489, opened 2022-04-26
UPDATED: qrUrl on provider to be optional

jeongmincho:
non-optional qrUrl made breaking change on extension, making the parameter optional.

### DIFF
--- a/src/provider/CoinbaseWalletProvider.ts
+++ b/src/provider/CoinbaseWalletProvider.ts
@@ -50,7 +50,7 @@ const HAS_CHAIN_OVERRIDDEN_FROM_RELAY = "HasChainOverriddenFromRelay";
 export interface CoinbaseWalletProviderOptions {
   chainId?: number;
   jsonRpcUrl: string;
-  qrUrl: string | null;
+  qrUrl?: string | null;
   overrideIsCoinbaseWallet?: boolean;
   overrideIsMetaMask: boolean;
   relayEventManager: WalletSDKRelayEventManager;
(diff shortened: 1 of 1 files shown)


src/provider/CoinbaseWalletProvider.ts
```diff
@@ -50,7 +50,7 @@ const HAS_CHAIN_OVERRIDDEN_FROM_RELAY = "HasChainOverriddenFromRelay";
 export interface CoinbaseWalletProviderOptions {
   chainId?: number;
   jsonRpcUrl: string;
-  qrUrl: string | null;
+  qrUrl?: string | null;
```
bangtoven, line 53: isn’t it supposed to be something like
```suggestion
  qrUrl?: string = undefined;
```
and come as last param or something?

bangtoven (commented):
@jeongmincho thanks for quick turnaround. one question on js syntax as i am a js noob

cb-jake (dismissed):

cb-heimdall:
Review Error for cb-jake @ 2022-04-26 23:47:49 UTC

src/provider/CoinbaseWalletProvider.ts
bangtoven: nvm. confirmed it works. i think i was confused with optional function param.

bangtoven (approved):

src/provider/CoinbaseWalletProvider.ts
erin-at-work: Does it need `null` at all?

PR #492, opened 2022-04-27
Initial threshold

erin-at-work:
### *Summary*

- Adds initial threshold (currently 'util' files) that'll fail if additions fall below


jest.config.ts
```diff
@@ -41,8 +41,14 @@ export default {
...
-  // An object that configures minimum threshold enforcement for coverage results
-  // coverageThreshold: undefined,
+  // TODO: Increase threshold as additional tests are added
+  coverageThreshold: {
+    global: {
+      branches: 18.3,
+      functions: 9.05,
+      statements: 9.44
+    }
+  },
```
ljharb, line 51: my preferred alternative to enforcing coverage in jest, is to do it with codecov (which can merge coverage reports across multiple jobs).

erin-at-work: Definitely want to replace with something more robust. Having this in as a quick stopgap.

brianchitester (approved):

PR #495, opened 2022-04-28
Release script

erin-at-work:
### *Summary*

- Release script

image
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,7 +37,8 @@
     "lint:prettier": "prettier --check \"{src,__tests__}/**/*.(js|ts|tsx)\"",
     "lint:eslint": "eslint ./src --ext .ts,.tsx",
     "fix:eslint": "yarn lint:eslint --fix",
-    "fix:prettier": "prettier . --write"
+    "fix:prettier": "prettier . --write",
+    "release": "./scripts/release.sh"
   },
   "private": true,
   "dependencies": {
--- a/package.json
+++ b/package.json
@@ -38,7 +38,7 @@
     "lint:eslint": "eslint ./src --ext .ts,.tsx",
     "fix:eslint": "yarn lint:eslint --fix",
     "fix:prettier": "prettier . --write",
-    "release": "./scripts/release.sh"
+    "prepublishOnly": "./scripts/prepublishOnly.sh"
   },
   "private": true,
   "dependencies": {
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -36,7 +36,8 @@
     "lint:prettier": "prettier --check \"{src,__tests__}/**/*.(js|ts|tsx)\"",
     "lint:eslint": "eslint ./src --ext .ts,.tsx",
     "fix:eslint": "yarn lint:eslint --fix",
-    "fix:prettier": "prettier . --write"
+    "fix:prettier": "prettier . --write",
+    "release": "./scripts/release.sh"
```
bangtoven, line 40: awesome!
can this be prepublish? https://yarnpkg.com/package/prepublish
not sure there’s post publish, but slack notifier could go there if exist.

scripts/release.sh
```diff
@@ -0,0 +1,47 @@
...
+  rm -rf ./node_modules
+  echo -e "yarn install"
+  yarn install
+  echo -e "yarn build:prod"
+  yarn build:prod
+  echo -e "cd build/npm"
+  cd build/npm
+  echo -e "-------------------------------------------------"
+  echo -e "npm publish"
+  npm publish
```
bangtoven, line 35: how about adding tag here, like release-candidate or next or beta, so that we can run this on ci/cd? then promote it as latest release manually?

package.json
ljharb: not sure why we'd use a deprecated package (or why you'd link to yarn instead of npm) but yes, this should be a `postpublish` script if it's meant to run after a manual publish.

however, if this is meant to replace *publishing*, then we probably want to add a `prepublishOnly` script that files if the right env var isn't set, and tells the user to run `yarn release`.

scripts/release.sh
```diff
@@ -0,0 +1,47 @@
...
+gitMessage=$(git log --oneline -n 1)
+mainBranch="master"
+branch=$(git rev-parse --abbrev-ref HEAD)
+
+echo -e "${PURPLE}Checking all branches are up-to-date..."
+echo -e "================================================="
+echo -e " git fetch --all"
+git fetch --all
+echo -e " git pull origin ${mainBranch}"
+git pull origin $mainBranch
```
ljharb, line 19: not sure if pull is what's desired here, since a) that does a fetch that's already done, and b) that merges, when we probably just want `git checkout origin/$mainBranch` instead of being on a branch at all?

cigis123:
Master

package.json
erin-at-work: Because I can't `cd` into a directory within a script, I won't be able to update this to `prepublishOnly`

vishnumad (approved):

bangtoven (approved):

PR #507, opened 2022-05-06
[chore] LinkFlow test

erin-at-work:
### _Summary_

image

### DIFF
--- a/src/CoinbaseWalletSDK.ts
+++ b/src/CoinbaseWalletSDK.ts
@@ -61,6 +61,7 @@ export class CoinbaseWalletSDK {
     if (!options.uiConstructor) {
       uiConstructor = opts => new WalletSDKUI(opts);
     } else {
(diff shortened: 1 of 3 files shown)


src/CoinbaseWalletSDK.ts
```diff
@@ -61,6 +61,7 @@ export class CoinbaseWalletSDK {
     if (!options.uiConstructor) {
       uiConstructor = opts => new WalletSDKUI(opts);
     } else {
+      /** @deprecated */
       uiConstructor = options.uiConstructor;
```
erin-at-work, line 65: Doesn't seem to referenced/used anywhere.

cb-jake: I think the extension uses this.

cb-jake (approved):

cb-jake (approved):

PR #517, opened 2022-05-11
expose session config

0xcaff:
### _Summary_

Allow session config to be exposed to users of WalletSDKRelay

### DIFF
--- a/src/relay/WalletSDKRelay.ts
+++ b/src/relay/WalletSDKRelay.ts
@@ -37,7 +37,6 @@ import {
   APP_VERSION_KEY,
   CancelablePromise,
   LOCAL_STORAGE_ADDRESSES_KEY,
-  SESSION_CONFIG_KEY,
   WALLET_USER_NAME_KEY,
   WalletSDKRelayAbstract,
 } from "./WalletSDKRelayAbstract";
@@ -128,11 +127,6 @@ export class WalletSDKRelay extends WalletSDKRelayAbstract {
     this.subscriptions.add(
       this.connection.sessionConfig$.subscribe({
         next: sessionConfig => {
-          this.storage.setItem(
-            SESSION_CONFIG_KEY,
-            JSON.stringify(sessionConfig),
-          );
-
           options.sessionConfigCallback?.(sessionConfig);
         },
         error: () => {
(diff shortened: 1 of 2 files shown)


src/relay/WalletSDKRelay.ts
```diff
@@ -122,6 +125,24 @@ export class WalletSDKRelay extends WalletSDKRelayAbstract {
       this.eventListener,
     );
 
+    this.subscriptions.add(
+      this.connection.sessionConfig$.subscribe({
+        next: sessionConfig => {
+          this.storage.setItem(
+            SESSION_CONFIG_KEY,
```
0xcaff, line 132: I don't love the idea of storing this but it seems like we expose quite a bit of info to caller pages today.

This doesn't expand our risk profile given we expose so much.

Screen Shot 2022-05-11 at 1 27 58 PM

etherealbeimg: I wish I understood all this stuff

src/relay/WalletSDKRelay.ts
```diff
@@ -1166,4 +1180,6 @@ export class WalletSDKRelay extends WalletSDKRelayAbstract {
         break;
     }
   }
+
+  protected onSessionConfigChanged(_nextSessionConfig: SessionConfig): void {}
```
hieronymus777, line 1184: Need to add this to `WalletLinkSDKRelayAbstract` https://github.com/coinbase/coinbase-wallet-sdk/blob/master/src/relay/WalletSDKRelayAbstract.ts

0xcaff: why?

hieronymus777: That's the class that the extension overrides.

hieronymus777 (approved):

src/relay/WalletSDKRelay.ts
```diff
@@ -122,6 +123,19 @@ export class WalletSDKRelay extends WalletSDKRelayAbstract {
       this.eventListener,
     );
 
+    this.subscriptions.add(
+      this.connection.sessionConfig$.subscribe({
+        next: sessionConfig => {
+          this.onSessionConfigChanged(sessionConfig);
```
brianchitester, line 129: Do you need to optionally call this function in the case it's not implemented?

0xcaff: This function will never not be implemented. It will either have a nop-empty implementation or one overridden by extending the class.

PR #518, opened 2022-05-11
[chore] RxWebSocket & WalletSDKRelayEventManager

erin-at-work:
### _Summary_

- Update node support versions
- Increase code coverage threshold
- Tests for:
  - WalletSDKRelayEventManager
  - RxWebSocket
image
image

- just tests


package.json
```diff
@@ -107,6 +108,6 @@
     "whatwg-fetch": "^3.5.0"
   },
   "engines": {
-    "node": ">= 10.0.0"
+    "node": ">= 14.0.0"
```
cb-jake, line 111: Would this constitute a major breaking change?

cb-jake (approved):
LGTM, I know I suggested the engines update but I'm curious if we should update it since it could be a breaking change for other apps.

package.json
erin-at-work: It's likely, 🤔  
cc @ljharb

cb-jake (approved):

package.json
ljharb: Yes, absolutely and always. why is this needed?

PR #554, opened 2022-06-06
Add browser-compatible buffer and util for Webpack 5

jmhacks:
### _Summary_
Added fallback for buffer and util node libraries for breaking dapps that are bundled with Webpack 5 build tools

Locally built with CRA dapp

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,6 +45,7 @@
     "@metamask/safe-event-emitter": "2.0.0",
     "bind-decorator": "^1.0.11",
     "bn.js": "^5.1.1",
+    "buffer": "^6.0.3",
     "clsx": "^1.1.0",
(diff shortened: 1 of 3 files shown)


cb-jake (approved):

brianchitester (approved):
tested in test dapp and it fixes the `Buffer is not defined` error i was seeing before

erin-at-work (approved):

jm42:
Doesn't work for me. Throws error at https://github.com/coinbase/coinbase-wallet-sdk/blob/ae895de54aa242eda39a11f8f205589256ad04a1/packages/wallet-sdk/src/relay/walletlink/ui/components/QRCode.tsx#L33

Isn't there a base64 util for browsers instead of using Node polyfill?

PR #561, opened 2022-06-08
Add Select Provider Method to CoinbaseWalletProvider

jeongmincho:
### _Summary_
Problem: When a user with both CBW and MM installed attempts to connect with a dapp via the injected provider, the user sees two separate connection pop-ups (one from each wallet). This method will enable CBW to handle that case more gracefully by showing a selection pop-up to choose between the two wallets.

NOTE: A feature for dapp developers to completely disable CBW as an injected provider will be accessible from a follow-up PR. This will be useful when a dapp has already integrated CBW as a separate option on their multi-wallet modal.

Unit test, locally with the following dapps:
- Uniswap
- Compound
- Zapper
- Opensea
- Curve
- Sushiswap
- AAVE
- Balancer
- Prettier-Test-Dapp


gareys-cb (approved):

cb-heimdall:
Review Error for gareys-cb @ 2022-06-14 19:43:11 UTC

cb-jake (approved):

PR #621, opened 2022-08-11
Handle addressChanged Event

jakepanitz:
### _Summary_

Added handling for a WalletLinkMessage of type "addressChanged" which calls _setAddresses() with the provided address. This allows a consumer of the provider to change addresses on a dapp without the dapp needing to first request the address.

Manually tested and confirmed:

- Behavior related to initial connection to a dapp remains unchanged.
- Upon receiving the new "addressChanged" event, _setAddresses is called and the address is updated in the provider
- If the dapp supports the "accountsChanged" event, the address will immediately update on the dapp. If not, the page will need to be refreshed before the new address is visible.

### DIFF
--- a/packages/wallet-sdk/src/provider/CoinbaseWalletProvider.test.ts
+++ b/packages/wallet-sdk/src/provider/CoinbaseWalletProvider.test.ts
@@ -1,3 +1,5 @@
+import { fireEvent } from "@testing-library/preact";
+
 import { MockRelayClass } from "../__mocks__/relay";
 import {
   MOCK_ADDERESS,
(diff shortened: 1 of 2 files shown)


vishnumad (approved):

szczurek1987 (commented):
- [
 ]

cb-jake (approved):
Changes look good. Here is a test you can add for the new event

```ts
  it("updates the providers address on a postMessage's "addressChanged" event", () => {
    const provider = setupCoinbaseWalletProvider();

    fireEvent(
      window,
      new MessageEvent("message", {
        data: { data: { action: "addressChanged", address: "0x0000000000000000000000000000000000001010" }, type: 'walletLinkMessage' },
        origin: "dapp.finance"
      })
    )

    // @ts-expect-error _addresses is private
    expect(provider._addresses).toEqual(['0x0000000000000000000000000000000000001010']);
  });
```

erin-at-work (dismissed):

jeongmincho (approved):

cb-jake (approved):

PR #623, opened 2022-08-12
Add bridge to SolanaProvider

willn-cb:
### _Summary_
This PR adds support for an injected bridge to the SolanaProvider.

If the bridge is present, it sends messages over the bridge's `postMessage` function rather than `window.postMessage`

Additionally adds support for a `WEB3_RESPONSE` response type, which will come from sources other than the Extension.

- Manually
- Unit tests

### DIFF
--- a/packages/wallet-sdk/src/provider/SolanaProvider.test.ts
+++ b/packages/wallet-sdk/src/provider/SolanaProvider.test.ts
@@ -67,6 +67,8 @@ describe("Solana Provider", () => {
 
   afterEach(() => {
     scopedStorageCopy.clear();
+    // @ts-expect-error need this to be null for most tests
+    window.__CIPHER_BRIDGE__ = null;
   });
 
   it("should post request and handle response for connect", async () => {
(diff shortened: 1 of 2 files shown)


cb-jake (commented):
Is the intent of this to work in the dapp browser or with walletlink? On first glance, seems like this is dapp browser specifc, in which case it should maybe live in cipher-web-view vs the sdk

willn-cb:
@cb-jake Yeah, the changes here are specific to the dapp browser of the mobile app. The SolanaProvider was added to the SDK for two reasons: 1. to be able to use the same code in both Extension and the mobile app and 2. to be able to add support for walletlink down the line.

This updated SolanaProvider will be imported into cipher-web-view so that it can be injected into the dapp browser along with the existing providers.

vishnumad (approved):

cb-jake (approved):

Tomengo18:
7c64ca65e57fe51f5214317be7ebf2b1f797ba2d

PR #627, opened 2022-08-12
remodel network switching

fan-zhang-sv:
### _Summary_
When making web3 provider, dapp will pass in a chain id as the default chain id for this dapp. SDK used to receive this chainId and still overriding it with a global user chosen default chainId. This PR changed this behaviour and utilize this dapp default chainId so user would be on dapp's default network when first connecting.

- added dapp's default network into walletlink qrcode
- removed hasChainChanged, etc. (simplified the logic of switching chain, if there's a call, there's a chain switch)

Unit Test
Manually Tested


shreypdev (dismissed):
Good to go 🚀

cb-heimdall:
Review Error for shreypdev @ 2022-08-25 19:19:37 UTC

erin-at-work (approved):

01SMD (commented):
.

PR #636, opened 2022-08-17
provide Account and chainId on the handshake response callback

bangtoven:


packages/wallet-native-sdk/ios/CoinbaseWalletSDK/Message/Request/Account.swift
```diff
@@ -9,10 +9,10 @@ import Foundation
 
 public struct Account: Codable {
     public let chain: String
-    public let networkId: UInt?
+    public let networkId: UInt
```
bangtoven, line 12: reverting https://github.com/coinbase/coinbase-wallet-sdk/commit/bbedddead2904ec68e36976008018e055be437ff#diff-4f9304d703a639926bc9f378947507e50f237bc4da9d801b23665741016b4b2e

AMITGOELNY (approved):

PR #710, opened 2022-10-27
[WALL-12550]: Updates check to ensure event origin matches location origin

AMITGOELNY:
### _Summary_
Updates check to ensure event origin matches location origin

manually


cb-jake (commented):
lgtm. We should add a test case for the first check in the if block.

AMITGOELNY:
> lgtm. We should add a test case for the first check in the if block.

Test case added and broken tests resolved

erin-at-work (approved):

Riani837:
Master

01SMD (commented):
.

PR #712, opened 2022-11-01
Updated DescriptionItem component so img tags are more accessible

riklomas:
### _Summary_

Added alt="" and aria-hidden="true" to icon images

Tested locally


bangtoven:
it seems TryExtensionLinkDialog.tsx file has been replaced with your change @cb-jake . do you think we should address the changes suggested here or can we close this?

PR #739, opened 2022-11-29
Enable webpack tree shaking

dawsbot:
https://webpack.js.org/guides/tree-shaking/

### _Summary_

Enable tree shaking by informing webpack that this package is `sideEffect` `false`

Still installs and compliles


bangtoven:
thanks for opening this PR @dawsbot . sorry for the late response on this.

image
@cb-jake @erin-at-work we have code to set properties of window, e.g. `window.CoinbaseWalletSDK` and there are some css files as well. do you think those should be specified as sideeffect?

PR #743, opened 2022-12-01
[Bug Fixed] Insufficient validation when decoding a `Coinbase-Socket` packet

iot-defcon:
👾 **Describe The Sumarry:** 
Due to improper type validation in the `socket.coinbase-parser` library (which is used by the socket.io and `socket.io-client` packages to encode and decode Socket.IO packets), it is possible to overwrite the _placeholder object which allows an attacker to place references to functions at arbitrary places in the resulting query object.

🪰 **Proof On Concepts:** 
Bellow the vulnerable of code's:
```js
const decoder = new Decoder();

decoder.on("decoded", (packet) => {
 console.log(packet.data); // prints [ 'hello', [Function: splice] ]
})

decoder.add('51-["coinbase",{"_placeholder":true,"num":"splice"}]');
decoder.add(Buffer.from("world"));
```
This bubbles up in the socket package:
```js
io.on("connection", (socket) => {
 socket.on("hello", (val) => {
 // here, "val" could be a function instead of a buffer
 });
});
```
You need to make sure that the payload that you received from the client is actually a Buffer object:
```js
io.on("connection", (socket) => {
 socket.on("coinbase", (val) => {
 if (!Buffer.isBuffer(val)) {
 socket.disconnect();
 return;
 }
 // ...
 });
});
```
vulnerable of values that could be sent by a malicious user:
 * a number that is out of bounds packet: `451-["hello",{"_placeholder":true,"num":10}]`

```js
io.on("connection", (socket) => {
 socket.on("coinbase", (val) => {
 // val is `undefined`
 });
});
```
 * a value that is not a number, like `undefined`

```
io.on("connection", (socket) => {
 socket.on("hello", (val) => {
 // val is `undefined`
 });
});
```

🛡️ **Type of change's:** 
`4.0.0` **>=**  `4.0.5`

🥷 **According CVeScores:** 
CVE-2022-2421
[CWE-20](https://cwe.mitre.org/data/definitions/20.html)
[CWE-1287](https://cwe.mitre.org/data/definitions/1287.html)
`CVSS:3.1/AV:N/AC:L/PR:N/UI:N/S:U/C:H/I:H/A:H`

### DIFF
--- a/packages/wallet-sdk/yarn.lock
+++ b/packages/wallet-sdk/yarn.lock
@@ -1609,6 +1609,7 @@
 "@types/component-emitter@^1.2.10":
   version "1.2.11"
   resolved "https://registry.yarnpkg.com/@types/component-emitter/-/component-emitter-1.2.11.tgz#50d47d42b347253817a39709fef03ce66a108506"
+  integrity sha512-SRXjM+tfsSlA9VuG8hGO2nft2p8zjXCK1VcC6N4NXbBbYbSia9kzCChYQajIjzIqOOOuh5Ock6MmV2oux4jDZQ==
 
 "@types/connect@^3.4.33":
   version "3.4.35"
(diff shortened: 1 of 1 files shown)


bangtoven:
thank you! addressed https://github.com/coinbase/coinbase-wallet-sdk/commit/5101d8e2584a4063dbf1645613e6267177250316

PR #778, opened 2023-01-09
AES256GCM decrypt: replace Observable with Promise

bangtoven:

### DIFF
--- a/packages/wallet-sdk/src/relay/aes256gcm.test.ts
+++ b/packages/wallet-sdk/src/relay/aes256gcm.test.ts
@@ -14,7 +14,7 @@ describe("aes256gcm", () => {
     // decrypted output matches original input
     decrypt(encrypted, randSecret).then(decrypted => {
       expect(decrypted).toBe("plain text string");
-    });
+    }, () => {});
   });
 
   test("decrypt", () => {
@@ -31,7 +31,7 @@ describe("aes256gcm", () => {
 
     decrypt(cipherText, secret).then(value => {
       expect(sampleDataResult).toEqual(value);
-    });
+    }, () => {});
   });
 
   test("errors", async () => {
--- a/packages/wallet-sdk/src/relay/aes256gcm.test.ts
+++ b/packages/wallet-sdk/src/relay/aes256gcm.test.ts
@@ -12,7 +12,7 @@ describe("aes256gcm", () => {
     expect(encrypted.length).toEqual(90);
 
     // decrypted output matches original input
-    decrypt(encrypted, randSecret).then(
+    return decrypt(encrypted, randSecret).then(
       decrypted => {
         expect(decrypted).toBe("plain text string");
       },
@@ -32,7 +32,7 @@ describe("aes256gcm", () => {
       },
     };
 
-    decrypt(cipherText, secret).then(
+    return decrypt(cipherText, secret).then(
       value => {
         expect(sampleDataResult).toEqual(JSON.parse(value));
       },
--- a/packages/wallet-sdk/src/relay/aes256gcm.test.ts
+++ b/packages/wallet-sdk/src/relay/aes256gcm.test.ts
@@ -12,10 +12,8 @@ describe("aes256gcm", () => {
     expect(encrypted.length).toEqual(90);
 
     // decrypted output matches original input
-    decrypt(encrypted, randSecret).subscribe({
-      next: decrypted => {
-        expect(decrypted).toBe("plain text string");
-      },
+    decrypt(encrypted, randSecret).then(decrypted => {
+      expect(decrypted).toBe("plain text string");
     });
   });
 
@@ -31,10 +29,8 @@ describe("aes256gcm", () => {
       },
     };
 
-    decrypt(cipherText, secret).subscribe({
-      next: value => {
-        expect(sampleDataResult).toEqual(value);
-      },
+    decrypt(cipherText, secret).then(value => {
+      expect(sampleDataResult).toEqual(value);
     });
   });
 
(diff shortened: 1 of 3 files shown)


packages/wallet-sdk/src/relay/aes256gcm.test.ts
```diff
@@ -31,11 +32,12 @@ describe("aes256gcm", () => {
       },
     };
 
-    decrypt(cipherText, secret).subscribe({
-      next: value => {
-        expect(sampleDataResult).toEqual(value);
+    decrypt(cipherText, secret).then(
+      value => {
+        expect(sampleDataResult).toEqual(JSON.parse(value));
```
bangtoven, line 37: wonder how this has been working
decrypt returns Promise (or used to be Observable) of string, so `value` (string) and `sampleDataResult` (object) couldn't be equal, right?
image

packages/wallet-sdk/src/relay/aes256gcm.test.ts
```diff
@@ -47,15 +49,6 @@ describe("aes256gcm", () => {
       "secret must be 256 bits",
     );
 
-    await expect(
-      decrypt("text", secret).subscribe(
-        () => {
-          fail("expected error");
-        },
-        error => {
-          expect(error).toBe("expected error");
```
bangtoven, line 56: not sure how this worked before as well.
`error`'s message is not equal to `expected error`

packages/wallet-sdk/src/relay/aes256gcm.test.ts
cb-jake: did the observable resolve to the string or was it an object?

packages/wallet-sdk/src/relay/aes256gcm.ts
```diff
@@ -63,15 +61,10 @@ export async function encrypt(
  * @param cipherText hex string representation of bytes in the order: initialization vector (iv),
  * auth tag, encrypted plaintext. IV is 12 bytes. Auth tag is 16 bytes.
  * @param secret hex string representation of 32-byte secret
- *
- * TODO: Update rxjs for promises
```
bangtoven, line 67: 👌

cb-jake (dismissed):
code changes look good to me.

packages/wallet-sdk/src/relay/aes256gcm.test.ts
bangtoven: cc @erin-at-work

packages/wallet-sdk/src/relay/aes256gcm.test.ts
vishnumad: I just logged out what the old test returns when run and it seems like it actually returns an object. Incorrect typing on the decrypt function?

image

bangtoven: @vishnumad i am getting a string with old code
image
image

mateuszficner746:
hell where?

packages/wallet-sdk/src/relay/aes256gcm.test.ts
vishnumad: @bangtoven Yup, you're correct it is a string. This test is just broken and will pass no matter what you do in the subscribe function because it's an observable. You can test this by replacing the current expect with `expect(true).toEqual(false);`. We need to actually block until the observable emits the value.

```js
 test("decrypt", done => {
    const cipherText =
      "06593325a922a928913b5c6ea26f848c4545bcea4e26c4f5ee745316ff22b2780aeccc565730514b2820a94b03f5f89fe7542a35bbdd87a1d52a4352f49482781113db09266c668696778e0a94bc9f866f1e92e7262fd0bb811838284cc64cbc4552b33e9c6fb2582cea4f49471d6d46a16a5c8ac83ee8483ed4dc01f1fde3bfd7a2f173715e0a8d09dd4907483f096a845bff698831ea277c1ca4223d3f6073174cb35119d0a795c1a9cb4f32ee1dcc254d8931";
    const sampleDataResult = {
      type: "WEB3_RESPONSE",
      id: "791fe0ec3dc3de49",
      response: {
        method: "requestEthereumAccounts",
        result: ["0xdf0635793e91d4f8e7426dbd9ed08471186f428d"],
      },
    };

    decrypt(cipherText, secret).subscribe({
      next: value => {
        expect(sampleDataResult).toEqual(value);
        done();
      },
    });
  });
```

vishnumad: @bangtoven You'll run into a similar issue with the promises in the current test. The test won't actually get marked as failed currently since you're not awaiting the result of the promise. 
image

```js
test("decrypt", async () => {
    // ...
    const value = await decrypt(cipherText, secret);
    expect(value).toEqual(sampleDataResult);
  });
```

bangtoven: @vishnumad hmm. actually i was able to find type error since the test failed.

vishnumad: @bangtoven It does print the error to console, but doesn't mark the test as failed. See the orange `RUNS` next to the last line in the screenshot above vs red `FAIL` below when using await
image

bangtoven: oh, gotcha. i just made change to the test code to return the promise.then block.

vishnumad (approved):

packages/wallet-sdk/src/relay/aes256gcm.test.ts
```diff
@@ -12,11 +12,12 @@ describe("aes256gcm", () => {
...
     // decrypted output matches original input
-    decrypt(encrypted, randSecret).subscribe({
-      next: decrypted => {
+    return decrypt(encrypted, randSecret).then(
+      decrypted => {
         expect(decrypted).toBe("plain text string");
       },
-    });
+      () => {},
+    );
```
cb-jake, line 20: nit: can we use async/await in tests? A bit easier to read (same below)

bangtoven: yup sure.

vishnumad (approved):

PR #319, opened 2022-02-09
Delete outdated WL server and webpage source code

bangtoven:
before
Screen Shot 2022-02-09 at 2 51 35 PM

after
Screen Shot 2022-02-09 at 2 51 04 PM

PR #779, opened 2023-01-09
Remove unused RxJS calls

bangtoven:

### DIFF
--- a/packages/wallet-sdk/src/relay/Session.ts
+++ b/packages/wallet-sdk/src/relay/Session.ts
@@ -1,8 +1,6 @@
 // Copyright (c) 2018-2022 Coinbase, Inc. <https://www.coinbase.com/>
 // Licensed under the Apache License, version 2.0
 
-import { fromEvent, Observable } from "rxjs";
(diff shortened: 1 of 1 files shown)


packages/wallet-sdk/src/relay/Session.ts
```diff
@@ -48,19 +46,6 @@ export class Session {
     return null;
   }
 
-  public static get persistedSessionIdChange$(): Observable<{
```
cb-jake, line 51: what was this originally used for?

bangtoven: @cb-jake i couldn't find any usages

cb-jake: Sounds good, maybe @hieronymus777 has context?

bangtoven: i did some digging, and it turned out this could have been deprecated along with this PR https://github.com/coinbase/coinbase-wallet-sdk/pull/319 
this is a dead code

cb-jake (approved):

PR #839, opened 2023-02-27
Update `ethereum` type to `any`

erin-at-work:
### _Summary_

Update `ethereum` to `any`

From issue: https://github.com/coinbase/coinbase-wallet-sdk/issues/343

### DIFF
--- a/packages/wallet-sdk/src/index.ts
+++ b/packages/wallet-sdk/src/index.ts
@@ -15,7 +15,7 @@ declare global {
     /**
      * For CoinbaseWalletSDK, window.ethereum is `CoinbaseWalletProvider`
      */
-    ethereum?: unknown;
+    ethereum?: any;
     coinbaseWalletExtension?: CoinbaseWalletProvider;
 
     /**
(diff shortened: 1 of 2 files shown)


karthedew:
Does line 240 need to be change?

 from:
`const ethereum = (window.ethereum ?? window.top?.ethereum) as any;` 

to:
`const ethereum = (window as any).ethereum ?? (window as any).top?.ethereum;`

packages/wallet-sdk/src/index.ts
```diff
@@ -15,7 +15,7 @@ declare global {
     /**
      * For CoinbaseWalletSDK, window.ethereum is `CoinbaseWalletProvider`
      */
-    ethereum?: unknown;
+    ethereum?: any;
```
cb-jake, line 18: why are we using any here? Seems like we should be using the provider `CoinbaseWalletProvider`

cb-jake: Thank you

cb-jake (approved):

PR #904, opened 2023-06-24
MobileRelayUI

bangtoven:


packages/wallet-sdk/src/CoinbaseWalletSDK.ts
```diff
@@ -45,7 +46,7 @@ export interface CoinbaseWalletSDKOptions {
   /** @optional whether or not to reload dapp automatically after disconnect, defaults to true */
   reloadOnDisconnect?: boolean;
   /** @optional whether to connect mobile web app via WalletLink, defaults to false */
-  useMobileWalletLink?: boolean;
+  enableMobileWalletLink?: boolean;
```
bangtoven, line 49: avoid confusion as react hook

packages/wallet-sdk/src/provider/MobileRelayUI.ts
```diff
@@ -0,0 +1,157 @@
...
+
+    window.open(url.href, "_blank");
+  }
+
+  requestEthereumAccounts(_options: {
+    onCancel: ErrorHandler;
+    onAccounts?: ((accounts: [AddressString]) => void) | undefined;
+  }): void {
+    const wl_url = "TODO";
+    this.openCoinbaseWalletDeeplink({ wl_url });
```
bangtoven, line 39: TODO

packages/wallet-sdk/src/relay/MobileRelay.ts
```diff
@@ -1,41 +1,73 @@
...
+  ): Observable<string> {
+    const observable = super.publishEvent(event, message, callWebhook);
+
+    if (
+      this._enableMobileWalletLink &&
+      event === "Web3Request" &&
+      message.type === RelayMessageType.WEB3_REQUEST
+    ) {
+      const web3RequestMessage = message as Web3RequestMessage;
+      this.openCoinbaseWalletDeeplink(web3RequestMessage.request);
```
bangtoven, line 52: note that there might be a potential edge case in this code. 
we have overridden the publishEvent function to open the cbwallet app whenever a client sends web3 requests to walletlink. (filtering out SESSION_ID_REQUEST and LINKED messages) ideally, this logic should appended to the super.observable.
however, we have encountered an issue where the browser seems to restrict calling window.open within the rxjs subscription block. so it's called when the observable is created.

```
      return observable.pipe(
        tap(() => {
          console.log("MobileRelay publishEvent tap within tap");
          this.openCoinbaseWalletDeeplink(web3RequestMessage.request);
        }),
      );

```

packages/wallet-sdk/src/relay/MobileRelay.ts
```diff
@@ -1,41 +1,58 @@
...
   }
+
+  // override
+  protected publishWeb3RequestEvent(id: string, request: Web3Request): void {
+    super.publishWeb3RequestEvent(id, request);
+
+    if (this._enableMobileWalletLink) {
+      this.openCoinbaseWalletDeeplink(request);
+    }
+  }
```
bangtoven, line 41: publishWeb3RequestEvent function has been overridden to open the cbwallet app when client sends web3 requests to walletlink. 

note that there might be a potential edge case in this code.
ideally, this logic should appended to publishEvent like below, being piped after publishEvent observable to get triggered when ws communication happens:
``` typescript
protected publishEvent(
    event: string,
    message: RelayMessage,
    callWebhook: boolean,
  ): Observable<string> {
    const observable = super.publishEvent(event, message, callWebhook);

    if (
      this._enableMobileWalletLink &&
      event === "Web3Request" &&
      message.type === RelayMessageType.WEB3_REQUEST
    ) {
      return observable.pipe(
        tap(() => {
          const web3RequestMessage = message as Web3RequestMessage;
          this.openCoinbaseWalletDeeplink(web3RequestMessage.request);
        }),
      );
    }

    return observable;
  }
```
however, we have encountered an issue where the browser seems to restrict calling window.open within the rxjs subscription block. 

if it can't be appended to the original observable, there's no point to modify it, since it's called with SESSION_ID_REQUEST and LINKED messages as well. (one of previous attempt below)
image 

so to simplify the logic, overriding `publishWeb3RequestEvent` instead.

packages/wallet-sdk/src/provider/MobileRelayUI.ts
```diff
@@ -0,0 +1,125 @@
...
+  EthereumAddressFromSignedMessageResponse,
+  SignEthereumMessageResponse,
+  SignEthereumTransactionResponse,
+  SubmitEthereumTransactionResponse,
+} from "../relay/Web3Response";
+import { AddressString, ProviderType } from "../types";
+import { WalletUI, WalletUIOptions } from "./WalletUI";
+
+// TODO: Implement & present in-page wallet picker instead of navigating to www.coinbase.com/connect-dapp
+export class MobileRelayUI implements WalletUI {
```
bangtoven, line 18: empty class conforming to `WalletUI`, which skips the creation of QR code dialogs and avoids running any business logic associated with desktop connection cases.

packages/wallet-sdk/src/provider/MobileRelayUI.ts
```diff
@@ -0,0 +1,125 @@
...
+
+  setConnectDisabled(_: boolean) {} // no-op
+
+  showConnecting(_options: {
+    isUnlinkedErrorState?: boolean | undefined;
+    onCancel: ErrorHandler;
+    onResetConnection: () => void;
+  }): () => void {
+    return () => {}; // no-op
+  }
```
bangtoven, line 104: TODO: i might try this thing

packages/wallet-sdk/src/relay/MobileRelay.ts
```diff
@@ -1,41 +1,60 @@
...
         location.href = `https://www.coinbase.com/connect-dapp?uri=${encodeURIComponent(
           location.href,
         )}`;
       }),
       cancel: () => {},
     };
   }
+
+  // override
+  protected publishWeb3RequestEvent(id: string, request: Web3Request): void {
```
bangtoven, line 35: publishWeb3RequestEvent function has been overridden to open the cbwallet app when client sends web3 requests to walletlink. 

note that there might be a potential edge case in this code.
ideally, this logic should appended to publishEvent's observable like below, being piped after original publishEvent as `tap` side effect to get triggered when ws message actually gets sent:
``` typescript
protected publishEvent(
    event: string,
    message: RelayMessage,
    callWebhook: boolean,
  ): Observable<string> {
    const observable = super.publishEvent(event, message, callWebhook);

    if (
      this._enableMobileWalletLink &&
      event === "Web3Request" &&
      message.type === RelayMessageType.WEB3_REQUEST
    ) {
      return observable.pipe(
        tap(() => {
          const web3RequestMessage = message as Web3RequestMessage;
          this.openCoinbaseWalletDeeplink(web3RequestMessage.request);
        }),
      );
    }

    return observable;
  }
```
however, have encountered an issue where the browser seems to restrict calling window.open within the rxjs subscription block. 
per some SO posts, it's known that browser blocks web apps to call window.open from rxjs handlers.

if it can't be appended to the original observable, there's no point to modify it, since it's called with SESSION_ID_REQUEST and LINKED messages as well. (one of previous attempt below)
image 

so to simplify the logic, overriding `publishWeb3RequestEvent` instead.

bangtoven, line 35: publishWeb3RequestEvent function has been overridden to open the cbwallet app when client sends web3 requests to walletlink. 

note that there might be a potential edge case in this code.
ideally, this logic should appended to publishEvent's observable like below, being piped after original publishEvent as `tap` side effect to get triggered when ws message actually gets sent:
``` typescript
protected publishEvent(
    event: string,
    message: RelayMessage,
    callWebhook: boolean,
  ): Observable<string> {
    const observable = super.publishEvent(event, message, callWebhook);

    if (
      this._enableMobileWalletLink &&
      event === "Web3Request" &&
      message.type === RelayMessageType.WEB3_REQUEST
    ) {
      return observable.pipe(
        tap(() => {
          const web3RequestMessage = message as Web3RequestMessage;
          this.openCoinbaseWalletDeeplink(web3RequestMessage.request);
        }),
      );
    }

    return observable;
  }
```
however, have encountered an issue where the browser seems to restrict calling window.open within the rxjs subscription block. 
per some SO posts, it's known that browser blocks web apps to call window.open from rxjs handlers.

if it can't be appended to the original observable, there's no point to modify it, since it's called with SESSION_ID_REQUEST and LINKED messages as well. (one of previous attempt below)
image 

so to simplify the logic, overriding `publishWeb3RequestEvent` instead.

packages/wallet-sdk/src/components/RedirectDialog/RedirectDialog.tsx
```diff
@@ -0,0 +1,67 @@
...
+import { SnackbarContainer } from "../Snackbar";
+import css from "./RedirectDialog-css";
+
+type RedirectDialogProps = {
+  title: string;
+  buttonText: string;
+  onButtonClick: () => void;
+};
+
+export class RedirectDialog {
```
bangtoven, line 14: to show a button in case of redirection getting blocked by browser
image

packages/wallet-sdk/src/provider/MobileRelayUI.ts
```diff
@@ -0,0 +1,170 @@
...
+      title: "Redirecting to Coinbase Wallet...",
+      buttonText: "Open",
+      onButtonClick: () => {
+        this.redirectToCoinbaseWallet(walletLinkUrl);
+      },
+    });
+
+    setTimeout(() => {
+      this.redirectToCoinbaseWallet(walletLinkUrl);
+    }, 99);
```
bangtoven, line 57: have encountered an issue where the browser seems to restrict calling window.open within the rxjs subscription block. 
per some SO posts, it's known that browser blocks web apps to call window.open from rxjs handlers.

packages/wallet-sdk/src/relay/MobileRelay.ts
```diff
@@ -1,41 +1,55 @@
...
-          if (isInIFrame() && window.top) {
-            location = window.top.location;
-          } else {
-            location = window.location;
-          }
-        } catch (e) {
-          location = window.location;
-        }
-
+        const location = getLocation();
```
bangtoven, line 26: moved to util

packages/wallet-sdk/src/relay/WalletLinkRelay.ts
```diff
@@ -863,7 +863,7 @@ export class WalletLinkRelay extends WalletSDKRelayAbstract {
   // WIP
   // eslint-disable-next-line @typescript-eslint/ban-ts-comment
   // @ts-ignore
-  private connectAndSignIn(_params: {
+  private connectAndSignIn(params: {
```
bangtoven, line 866: addressing https://github.com/coinbase/coinbase-wallet-sdk/pull/908#discussion_r1245424509

packages/wallet-sdk/src/CoinbaseWalletSDK.ts
```diff
@@ -113,6 +108,11 @@ export class CoinbaseWalletSDK {
 
     this._relayEventManager = new WalletSDKRelayEventManager();
 
+    const isMobile = isMobileWeb();
+    const uiConstructor =
+      options.uiConstructor ||
+      (opts => new (isMobile ? MobileRelayUI : WalletLinkRelayUI)(opts));
```
cb-jake, line 114: while this works its kind of hard to read

```ts
(opts => new (isMobile ? MobileRelayUI : WalletLinkRelayUI)(opts));
```

the more verbose version is a bit easier to read

```ts
(opts => isMobile ? new MobileRelayUI(opts) : new WalletLinkRelayUI(opts);
```

packages/wallet-sdk/src/components/RedirectDialog/RedirectDialog.tsx
```diff
@@ -0,0 +1,67 @@
...
+  }
+}
+
+const RedirectDialogContent: FunctionComponent<
+  RedirectDialogProps & {
+    onDismiss: () => void;
+  }
+> = ({ title, buttonText, onButtonClick, onDismiss }) => {
+  return (
+    <SnackbarContainer darkMode={false}>
```
cb-jake, line 56: we should pass the darkmode flag vs hard coding

packages/wallet-sdk/src/relay/WalletLinkRelay.ts
```diff
@@ -863,7 +863,7 @@ export class WalletLinkRelay extends WalletSDKRelayAbstract {
   // WIP
   // eslint-disable-next-line @typescript-eslint/ban-ts-comment
   // @ts-ignore
```
cb-jake, line 865: Why is this ignored? have we not added connectAndSignIn to the relay abstract?

vishnumad (approved):

packages/wallet-sdk/src/CoinbaseWalletSDK.ts
bangtoven: thanks for the feedback. i wasn't aware what js convention on this was, so thought this version was easier to read. will make change.

packages/wallet-sdk/src/components/RedirectDialog/RedirectDialog.tsx
bangtoven: we don't have separate design themes for this redirect dialog, and it uses `SnackbarContainer` to make it as overlay dialog. the container doesn't seem like to have any UI in itself btw, so decided to pass hardcoded value here.

packages/wallet-sdk/src/relay/WalletLinkRelay.ts
bangtoven: it's not implemented yet.
it gives error since it's private and not being used.

packages/wallet-sdk/src/CoinbaseWalletSDK.ts
```diff
@@ -122,12 +122,10 @@ export class CoinbaseWalletSDK {
...
       diagnosticLogger: this._diagnosticLogger,
       reloadOnDisconnect: this._reloadOnDisconnect,
-      useMobileWalletLink: options.useMobileWalletLink,
+      enableMobileWalletLink: options.enableMobileWalletLink,
     };
 
-    this._relay = isMobileWeb()
-      ? new MobileRelay(relayOption)
-      : new WalletLinkRelay(relayOption);
+    this._relay = new (isMobile ? MobileRelay : WalletLinkRelay)(relayOption);
```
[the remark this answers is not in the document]
bangtoven: gotcha

packages/wallet-sdk/src/provider/MobileRelayUI.ts
```diff
@@ -0,0 +1,169 @@
...
+  private readonly redirectDialog: RedirectDialog;
+  private attached = false;
+
+  constructor() {
+    this.redirectDialog = new RedirectDialog();
+  }
+
+  attach() {
+    if (this.attached) {
+      throw new Error("Coinbase Wallet SDK UI is already attached");
```
cb-jake, line 29: should we actually throw here or is it better to just return early

packages/wallet-sdk/src/provider/MobileRelayUI.ts
cb-jake: is the goal here to auto redirect if the user doesn't click the button?

packages/wallet-sdk/src/components/RedirectDialog/RedirectDialog.scss
```diff
@@ -0,0 +1,53 @@
+@import "../theme.scss";
```
cb-jake, line 1: we should add dark mode support

packages/wallet-sdk/src/components/RedirectDialog/RedirectDialog.tsx
```diff
@@ -0,0 +1,67 @@
...
+      <div class={clsx("-cbwsdk-redirect-dialog")}>
```
cb-jake, line 57: clsx isn't needed here since we only have a single className

packages/wallet-sdk/src/components/RedirectDialog/RedirectDialog.tsx
```diff
@@ -0,0 +1,67 @@
...
+        <style>{css}</style>
+        <div class="-cbwsdk-redirect-dialog-backdrop" onClick={onDismiss} />
+        <div class="-cbwsdk-redirect-dialog-box">
+          <p>{title}</p>
+          <button onClick={onButtonClick}>{buttonText}</button>
```
cb-jake, line 62: do we have any designs to match existing sdk UI? The button here is a bit barebones

packages/wallet-sdk/src/components/RedirectDialog/RedirectDialog.tsx
```diff
@@ -0,0 +1,67 @@
...
+      <RedirectDialogContent
+        {...props}
+        onDismiss={() => {
+          this.clear();
+        }}
+      />,
+      this.root,
+    );
+  }
+}
```
cb-jake, line 48: This is confusing. we initially render null then check for props then return again

could we just do this

```suggestion
  private render(props: RedirectDialogProps | null): void {
    if (!this.root || !props) return;
    render(
      <RedirectDialogContent
        {...props}
        onDismiss={() => {
          this.clear();
        }}
      />,
      this.root,
    );
  }
}

```

packages/wallet-sdk/src/components/RedirectDialog/RedirectDialog.tsx
cb-jake: ahh that makes sense. We should address these UI inconsistencies during the refactor

packages/wallet-sdk/src/provider/MobileRelayUI.ts
bangtoven: it followed the same pattern as the existing class 
image

packages/wallet-sdk/src/components/RedirectDialog/RedirectDialog.tsx
bangtoven: applied the same button design as the `install` button on conection dialog linked to chrome store

packages/wallet-sdk/src/components/RedirectDialog/RedirectDialog.tsx
bangtoven: i found it needs to call `render(null, this.root);` here, otherwise sometimes it would stack multiple instances of this dialog.
note that, without proper refactor of WalletLinkRelay, it doesn't seem straightfoward to fine tune its `this.ui` behavior.

packages/wallet-sdk/src/components/RedirectDialog/RedirectDialog.scss
bangtoven: what would you recommend for dark mode? do you think background and font color would be enough?

btw, note that this method is supposed to be used by AMEX only first, and don't have plan to expose this externally.

packages/wallet-sdk/src/components/RedirectDialog/RedirectDialog.tsx
bangtoven: will address

packages/wallet-sdk/src/provider/MobileRelayUI.ts
cb-jake: cool that works then

packages/wallet-sdk/src/components/RedirectDialog/RedirectDialog.tsx
cb-jake: interesting. Okay to leave it for now but something we should address in future versions

packages/wallet-sdk/src/components/RedirectDialog/RedirectDialog.scss
cb-jake: Gotcha probably fine to leave for now

cb-jake (dismissed):

cb-jake (approved):

packages/wallet-sdk/src/components/RedirectDialog/RedirectDialog.scss
bangtoven: we can re visit coloring later (this project is for a hack so it doesn't have design support), but here's darkmode @cb-jake.
note that this dialog is only for an edge case where the browser blocks redirection.
Simulator Screenshot - iPhone 13 mini - 2023-06-29 at 13 23 45
Simulator Screenshot - iPhone 13 mini - 2023-06-29 at 13 20 25

PR #934, opened 2023-07-13
connect and sign: assert environments

bangtoven:

### DIFF
--- a/packages/wallet-sdk/src/mocks/relay.ts
+++ b/packages/wallet-sdk/src/mocks/relay.ts
@@ -5,7 +5,6 @@ import { CancelablePromise, WalletSDKRelayAbstract } from '../relay/WalletSDKRel
 import { Web3Method } from '../relay/Web3Method';
 import {
   AddEthereumChainResponse,
-  ConnectAndSignInResponse,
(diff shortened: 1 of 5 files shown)


packages/wallet-sdk/src/relay/WalletLinkRelay.ts
```diff
@@ -783,30 +781,6 @@ export class WalletLinkRelay extends WalletSDKRelayAbstract {
     }
   }
 
-  connectAndSignIn(params: {
```
bangtoven, line 786: moved to MobileRelay.ts

packages/wallet-sdk/src/mocks/relay.ts
```diff
@@ -143,10 +142,6 @@ export class MockRelayClass extends WalletSDKRelayAbstract {
     return;
   }
 
-  connectAndSignIn(): CancelablePromise<ConnectAndSignInResponse> {
```
bangtoven, line 146: remove `connectAndSignIn` from abstract class.
`connectAndSignIn` will be defined only on MobileRelay for now.

cb-jake (approved):

PR #940, opened 2023-07-15
fix: vite compatibility issue

### LINKED ISSUE #874: Bug: Incompatibility issue with vite
### Describe the bug

In vite, installed in a project with vue 2, the environment variables import.meta.env are used. As a result, after installing the module an error occurs in the file node_modules/@coinbase/wallet-sdk/dist/CoinbaseWalletSDK.js - "process is not defined", because by default the module uses variables of type process.env.

### Steps

1. In an existing project with vue 2 and @coinbase/wallet-sdk installed, install vite and remove vue-cli dependencies
2. Install vite compatibility plugin with vue 2: npm i @vitejs/plugin-vue2
3. adjust the vite.config.js file and replace all process.env with import.meta.env
3. Run the project: npm run dev

### Expected behavior

I would like to be able to override the process.env assignment within the plugin to the one supported in vite - import.meta.env. Ideally, if it were possible to configure this in the vite.config.js file

### Version

"@coinbase/wallet-sdk": "^3.5.2"

### Additional info

_No response_

### Desktop

_No response_

### Smartphone

_No response_

VladimirMikulic:
### _Summary_

Resolves #874

TSC is used to compile release version of the library, which can't polyfill/substitute process.env object like Webpack.
This caused compilation error since process.env doesn't exist in the browser (unless substituted).

### DIFF
--- a/packages/wallet-sdk/src/CoinbaseWalletSDK.ts
+++ b/packages/wallet-sdk/src/CoinbaseWalletSDK.ts
@@ -4,6 +4,7 @@
 import { LogoType, walletLogo } from './assets/wallet-logo';
 import { DiagnosticLogger } from './connection/DiagnosticLogger';
 import { EventListener } from './connection/EventListener';
+import { LINK_API_URL } from './constants';
 import { ScopedLocalStorage } from './lib/ScopedLocalStorage';
 import { CoinbaseWalletProvider } from './provider/CoinbaseWalletProvider';
 import { MobileRelayUI } from './provider/MobileRelayUI';
@@ -14,11 +15,6 @@ import { WalletLinkRelay } from './relay/WalletLinkRelay';
 import { WalletSDKRelayEventManager } from './relay/WalletSDKRelayEventManager';
 import { getFavicon, isMobileWeb } from './util';
 import { LIB_VERSION } from './version';
-
-const LINK_API_URL =
-  (typeof process !== 'undefined' && process.env?.LINK_API_URL) || 'https://www.walletlink.org';
-
-const SDK_VERSION = (typeof process !== 'undefined' && process.env?.SDK_VERSION) || LIB_VERSION;
 
 /** Coinbase Wallet SDK Constructor Options */
 export interface CoinbaseWalletSDKOptions {
@@ -52,7 +48,7 @@ export interface CoinbaseWalletSDKOptions {
 }
 
 export class CoinbaseWalletSDK {
-  public static VERSION = SDK_VERSION;
+  public static VERSION = LIB_VERSION;
 
   private _appName = '';
   private _appLogoUrl: string | null = null;
@@ -115,7 +111,7 @@ export class CoinbaseWalletSDK {
 
     const relayOption = {
       linkAPIUrl,
-      version: SDK_VERSION,
+      version: LIB_VERSION,
       darkMode: !!options.darkMode,
       uiConstructor,
       storage: this._storage,
--- a/packages/wallet-sdk/src/CoinbaseWalletSDK.ts
+++ b/packages/wallet-sdk/src/CoinbaseWalletSDK.ts
@@ -13,11 +13,12 @@ import { MobileRelay } from './relay/MobileRelay';
 import { WalletLinkRelay } from './relay/WalletLinkRelay';
 import { WalletSDKRelayEventManager } from './relay/WalletSDKRelayEventManager';
 import { getFavicon, isMobileWeb } from './util';
+import { LIB_VERSION } from './version';
 
-const LINK_API_URL = process.env.LINK_API_URL || 'https://www.walletlink.org';
+const LINK_API_URL =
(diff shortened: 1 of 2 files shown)


VladimirMikulic:
@cb-jake @bangtoven I'd appreciate a review when you're available. Thanks :)

packages/wallet-sdk/src/CoinbaseWalletSDK.ts
```diff
@@ -13,11 +13,12 @@ import { MobileRelay } from './relay/MobileRelay';
...
 import { getFavicon, isMobileWeb } from './util';
+import { LIB_VERSION } from './version';
 
-const LINK_API_URL = process.env.LINK_API_URL || 'https://www.walletlink.org';
+const LINK_API_URL =
+  (typeof process !== 'undefined' && process.env?.LINK_API_URL) || 'https://www.walletlink.org';
 
-// eslint-disable-next-line @typescript-eslint/no-var-requires
-const SDK_VERSION = process.env.SDK_VERSION! || require('../package.json').version || 'unknown';
+const SDK_VERSION = (typeof process !== 'undefined' && process.env?.SDK_VERSION) || LIB_VERSION;
```
cb-jake, line 21: We should just remove the process.env vars.

```suggestion
const SDK_VERSION = LIB_VERSION || require("./package.json").version; // will require a //@ts-expect-error or eslint disable.
```

packages/wallet-sdk/src/CoinbaseWalletSDK.ts
```diff
@@ -13,11 +13,12 @@ import { MobileRelay } from './relay/MobileRelay';
 import { WalletLinkRelay } from './relay/WalletLinkRelay';
 import { WalletSDKRelayEventManager } from './relay/WalletSDKRelayEventManager';
 import { getFavicon, isMobileWeb } from './util';
+import { LIB_VERSION } from './version';
 
-const LINK_API_URL = process.env.LINK_API_URL || 'https://www.walletlink.org';
+const LINK_API_URL =
+  (typeof process !== 'undefined' && process.env?.LINK_API_URL) || 'https://www.walletlink.org';
```
cb-jake, line 19: Lets make a constants file in the wallet-sdk package root and just define the `LINK_API_URL` var as a constant within that file and import that here.

cb-jake (changes requested):
see inline comments

cb-heimdall:
Review Error for cb-jake @ 2023-07-27 17:02:57 UTC

packages/wallet-sdk/src/CoinbaseWalletSDK.ts
VladimirMikulic: :+1:

packages/wallet-sdk/src/CoinbaseWalletSDK.ts
VladimirMikulic: :+1:

packages/wallet-sdk/src/CoinbaseWalletSDK.ts
VladimirMikulic: I replaced SDK_VERSION with LIB_VERSION since it's always defined in version.ts so require("./package.json") will never run.

VladimirMikulic:
Thank you @cb-jake! Done.

packages/wallet-sdk/src/CoinbaseWalletSDK.ts
```diff
@@ -51,7 +48,7 @@ export interface CoinbaseWalletSDKOptions {
 }
 
 export class CoinbaseWalletSDK {
-  public static VERSION = SDK_VERSION;
+  public static VERSION = LIB_VERSION;
```
cb-jake, line 51: I think we should still import the package json as a fallback. `require('../package.json').version` This should happen outside of the class

VladimirMikulic: @cb-jake I could import it, but it's unreachable code / it will never get referenced/used.
(LIB_VERSION is hard coded string in version.ts file which gets updated for each release, it's truthy value)

cb-jake: Aww, I was thinking the version file only existed when a build happens. Nvmd. This looks good as is. Thanks!

cb-jake:
Changes look good and this helps remove the need for the process polyfill. Once the build succeeds I'll get it merged

VladimirMikulic:
Amazing! Thank you so much @cb-jake!

packages/wallet-sdk/src/CoinbaseWalletSDK.ts
VladimirMikulic: All good :+1:

cb-jake (approved):

scotthconner:
This hasn't been released yet has it? Still ooks like 3.7.1 is the latest and that still has the issue.

PR #951, opened 2023-07-24
Remove safe-event-emitter in favor of eventemitter3

cb-jake:
### _Summary_

* Removed safe-event-emitter in favor of eventemitter3.
* While eventemitter3 is slightly bigger 500b v 1.1 kb it removes part of the need for the events polyfill (a future PR)

testapp and local dapp.


apps/testapp/src/context/CBWSDKProvider.tsx
```diff
@@ -16,10 +16,29 @@ export function CBWSDKProvider({ children }: CBWSDKProviderProps) {
     });
     setSdk(cbwsdk);
     const cbwprovider = cbwsdk.makeWeb3Provider('http');
-    console.info('provider', cbwprovider);
     setProvider(cbwprovider);
   }, []);
 
+  useEffect(() => {
+    if (provider) {
+      provider.on('connect', (info) => {
```
cb-jake, line 24: added the EIP-1193 events to see in the console when fired

packages/wallet-sdk/src/provider/SubscriptionManager.ts
```diff
@@ -23,11 +23,11 @@ export interface SubscriptionNotification {
...
 export class SubscriptionManager {
   private readonly subscriptionMiddleware: SubscriptionMiddleware;
-  readonly events: SafeEventEmitter;
+  readonly events: EventEmitter;
 
-  constructor(provider: Web3Provider & SafeEventEmitter) {
+  constructor(provider: Web3Provider & EventEmitter) {
     const blockTracker = new PollingBlockTracker({
-      provider: provider as Provider,
+      provider: provider as unknown as Provider,
```
bangtoven, line 30: why do we have `Web3Provider` AND `Provider` 😢

bangtoven (approved):

packages/wallet-sdk/src/provider/SubscriptionManager.ts
cb-jake: Provider is the type that comes from the `eth-block-tracker` package, which provides the `PollingBlockTracker` class

PR #953, opened 2023-07-26
pkg clean up

cb-jake:
### _Summary_

This was started to remove the need for URL polyfill but based on browser support changes are unneeded. We have 96%+ coverage across all major browsers. IE and Opera Mini are the only two without support. This is okay.

I've cleaned up a few unused scripts and updated tsconfigs.

image

Manually


packages/wallet-sdk/global.d.ts
```diff
@@ -0,0 +1,8 @@
+export {};
+declare global {
+  namespace jest {
+    interface Matchers<R> {
+      toThrowEIPError(code: number, message: string): R;
+    }
+  }
+}
```
bangtoven, line 8: do we need this even thought it is only for tests? @cb-jake

cb-jake: Yeah I think so. We're getting type errors in the test and our tests should be type safe as well

bangtoven (dismissed):

cb-heimdall:
Review Error for bangtoven @ 2023-07-26 20:14:34 UTC

packages/wallet-sdk/global.d.ts
bangtoven: gotcha. i think i added this to jest.setup.ts and it's been working so far
do we need to update anything on it?

bangtoven: @cb-jake it might be because `jest.setup.ts` got missed on tsconfig? 
image

cb-jake: Oh maybe it did get missed let me try that real quick

cb-jake: @bangtoven Looks like including the setup file fixed the types! New changes are pushed now

bangtoven (approved):

PR #954, opened 2023-07-27
Fix 4902 (unsupportedChain) error getting wrapped as -32603 & Fork eth-rpc-errors

bangtoven:


packages/wallet-sdk/src/errors/utils.ts
```diff
@@ -0,0 +1,150 @@
...
+      }
+    } else {
+      serialized.message = getMessageFromCode((serialized as SerializedEthereumRpcError).code);
+
+      serialized.data = { originalError: assignOriginalError(error) };
+    }
+  } else {
+    serialized.code = standardErrorCodes.rpc.internal;
+
+    const message = (error as any)?.message;
```
cb-jake, line 121: We should check the error using `instanceof` vs using any

```suggestion
let message = '';
if (error instanceof Error) {
 message =  error.message
}
```

packages/wallet-sdk/src/errors/utils.ts
```diff
@@ -0,0 +1,150 @@
...
+
+    serialized.message = message && typeof message === 'string' ? message : FALLBACK_MESSAGE;
+    serialized.data = { originalError: assignOriginalError(error) };
+  }
+
+  const stack = (error as any)?.stack;
```
cb-jake, line 127: We should follow the same pattern here

packages/wallet-sdk/src/errors/utils.ts
bangtoven: this is the code forked from eth-rpc-errors package.
btw, i don't think we should check whether it's Error instance. this code should be able to process any objects that has `message` prop even though they are not `Error`. 
is there better way to do that?

cb-jake: Hmm.. we could maybe use an unknown vs any

```ts
message = (error as unknown as { message?: string })?.message
```

or something else could be

```ts
func hasMsgProperty(obj: unknown): obj is {message: string} {
  return typeof obj === 'object' && obj !== null && 'message' in obj;
}

const message = hasMsgProperty(error) ? error.message : FALLBACK_MESSAGE
```

cb-jake: Option 2 is probably the best since it doesn't typecast

packages/wallet-sdk/src/errors/utils.ts
cb-jake: We could make this generic and use it for both cases

```ts
function hasProperty(obj: unknown, prop: string): obj is {stack: string} {
  return typeof obj === 'object' && obj !== null && prop in obj;
}

const stack = hasProperty(error, 'stack') ? error.stack : undefined
```

packages/wallet-sdk/src/errors/utils.ts
bangtoven: thanks for the examples!

packages/wallet-sdk/src/errors/utils.ts
```diff
@@ -118,16 +118,12 @@ export function serialize(
   } else {
     serialized.code = standardErrorCodes.rpc.internal;
 
-    const message = (error as any)?.message;
-
-    serialized.message = message && typeof message === 'string' ? message : FALLBACK_MESSAGE;
+    serialized.message = hasStringProperty(error, 'message') ? error.message : FALLBACK_MESSAGE;
```
bangtoven, line 121: thanks for the suggestions @cb-jake

cb-jake (approved):

PR #955, opened 2023-07-27
remove: 2 unused dependencies

VladimirMikulic:
### _Summary_

These dependencies were [previously used](https://github.com/coinbase/coinbase-wallet-sdk/pull/554/files) as polyfills in webpack bundle process but now that webpack is gone, they can be removed too.

`yarn test` succeeds and I confirmed they were not used in source code and only as part of webpack build [in the past](https://github.com/coinbase/coinbase-wallet-sdk/pull/554/files).

### DIFF
--- a/packages/wallet-sdk/package.json
+++ b/packages/wallet-sdk/package.json
@@ -46,9 +46,7 @@
     "keccak": "^3.0.3",
     "preact": "^10.16.0",
     "rxjs": "^6.6.3",
-    "sha.js": "^2.4.11",
-    "stream-browserify": "^3.0.0",
-    "util": "^0.12.5"
+    "sha.js": "^2.4.11"
   },
   "devDependencies": {
     "@babel/core": "^7.22.9",
--- a/yarn.lock
+++ b/yarn.lock
@@ -2715,13 +2715,11 @@ __metadata:
     rxjs: ^6.6.3
     sass: ^1.64.1
     sha.js: ^2.4.11
-    stream-browserify: ^3.0.0
     svgo: ^2.8.0
     ts-jest: ^27.1.5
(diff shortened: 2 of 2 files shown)


packages/wallet-sdk/package.json
```diff
@@ -36,7 +36,6 @@
     "@solana/web3.js": "^1.78.0",
     "bind-decorator": "^1.0.11",
     "bn.js": "^5.2.1",
-    "buffer": "^6.0.3",
```
cb-jake, line 39: Buffer doesn't exist in the browser. We'll be keeping this package and updating the code to reference it directly vs using the node builtin

packages/wallet-sdk/package.json
```diff
@@ -46,9 +45,7 @@
     "keccak": "^3.0.3",
     "preact": "^10.16.0",
     "rxjs": "^6.6.3",
-    "sha.js": "^2.4.11",
-    "stream-browserify": "^3.0.0",
```
cb-jake, line 50: Thanks for catching these two.

VladimirMikulic: You're welcome!

packages/wallet-sdk/package.json
VladimirMikulic: All current bundlers should do it by default now, but we can leave since you prefer it :+1:

VladimirMikulic:
@cb-jake comments resolved. Thanks!

cb-jake (approved):

cb-jake:
Going to wait till tests run and pass before merging

VladimirMikulic:
Thanks @cb-jake!

PR #957, opened 2023-07-28
fix dev

cb-jake:


packages/wallet-sdk/global.d.ts
```diff
@@ -0,0 +1,10 @@
+// This is not ideal. ts should pick up custom matchers from the jest.setup.ts file
+// but it doesn't. So we have to declare them here.
```
VladimirMikulic, line 2: Have you checked out https://jestjs.io/docs/expect#expectextendmatchers ?

bangtoven (dismissed):

erin-at-work (approved):

PR #963, opened 2023-08-03
[Bugfix] Read IsStandaloneSigning from Storage

fan-zhang-sv:
### _Summary_

CBW Extension used to read `IsStandaloneSigning` from storage and pass in this value to CoinbaseWalletProvider during its initialization. However, when user first open a dapp and before they connect wallet, there would be no `IsStandaloneSigning` saved in storage, and it caused issue for when user switches address. Changing it to read this value from storage directly when user switches address.

Manually tested on 
1. Standalone Extension: Dapp address switches as user switch address from extension
2. Walletlinked Extension: Dapp address does not switch as user switch address from RN
3. RN browser: Does not utilize this variable
4. RN Walletlink: Does not utilize this variable

### DIFF
--- a/packages/wallet-sdk/src/provider/CoinbaseWalletProvider.ts
+++ b/packages/wallet-sdk/src/provider/CoinbaseWalletProvider.ts
@@ -42,9 +42,10 @@ import {
 } from "./SubscriptionManager";
 import { RequestArguments, Web3Provider } from "./Web3Provider";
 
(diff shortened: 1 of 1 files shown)


cb-jake (approved):

bangtoven (changes requested):
let's see if we can achieve this without introducing sdk changes

PR #964, opened 2023-08-04
yet-another-testing-dapp

erin-at-work:
init app


bangtoven:
hey @erin-at-work do we plan to land this pr someday, or should we close it and revisit later?

erin-at-work:
> hey @erin-at-work do we plan to land this pr someday, or should we close it and revisit later?

@bangtoven ah, thanks for resurfacing. We should probably table this til we know whether this is an experience that the team wants to continue building on.

PR #968, opened 2023-08-07
chore: add `sideEffects: false` to package.json

tmm:
Adds `"sideEffects": false` to package.json for maximum tree-shakability.

Without `"sideEffects"` switched off build systems aren't able to properly eliminate dead code (e.g. Next.js) and could still load the Coinbase Wallet SDK JS even if it's not used.

More info: https://webpack.js.org/guides/tree-shaking/#mark-the-file-as-side-effect-free


bangtoven (approved):
thanks @tmm !

PR #975, opened 2023-08-11
Check unseen events after navigating back to browser

vishnumad:
### _Summary_
For mobile connections on iOS, sometimes Safari will kill the websocket connection in the background causing us to miss events. 

This hacky fix checks the `/events` endpoint for unseen events and merges it into the incoming events observable. Also needed to add some logic to re-connect the RxWebsocket connection instead of throwing an error when `sendData` is called while the websocket is connected. From testing, this seems to work fairly reliably, but a more robust solution would be better in the long term.

### DIFF
--- a/packages/wallet-sdk/src/connection/WalletLinkConnection.ts
+++ b/packages/wallet-sdk/src/connection/WalletLinkConnection.ts
@@ -211,16 +211,20 @@ export class WalletLinkConnection {
 
     // mark unseen events as seen
     this.subscriptions.add(
-      this.unseenEventsSubject.subscribe(async e => {
+      this.unseenEventsSubject.subscribe(e => {
         const credentials = `${this.sessionId}:${this.sessionKey}`;
         const auth = `Basic ${btoa(credentials)}`;
 
-        await fetch(`${this.linkAPIUrl}/events/${e.eventId}/seen`, {
-          method: "POST",
-          headers: {
-            Authorization: auth,
-          },
-        });
+        try {
+          fetch(`${this.linkAPIUrl}/events/${e.eventId}/seen`, {
+            method: "POST",
+            headers: {
+              Authorization: auth,
+            },
+          });
+        } catch (e) {
+          console.error("Unabled to mark event as failed:", e);
+        }
       }),
     );
   }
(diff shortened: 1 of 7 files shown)


bangtoven (approved):

packages/wallet-sdk/src/connection/WalletLinkConnection.ts
```diff
@@ -206,6 +208,25 @@ export class WalletLinkConnection {
           });
         }),
     );
+
+    // mark unseen events as seen
+    this.subscriptions.add(
+      this.unseenEventsSubject.subscribe(e => {
+        const credentials = `${this.sessionId}:${this.sessionKey}`;
+        const auth = `Basic ${btoa(credentials)}`;
```
bangtoven, line 216: optional: dedup this

packages/wallet-sdk/src/connection/WalletLinkConnection.ts
```diff
@@ -308,6 +329,47 @@ export class WalletLinkConnection {
...
+    const credentials = `${this.sessionId}:${this.sessionKey}`;
+    const auth = `Basic ${btoa(credentials)}`;
+
+    const response = await fetch(`${this.linkAPIUrl}/events?unseen=true`, {
+      headers: {
+        Authorization: auth,
+      },
+    });
+
+    if (response.ok) {
```
bangtoven, line 338: optional:
```suggestion

    if (!response.ok) {
      throw new Error(`Check unseen events failed: ${response.status}`);
    }
    
    
```

bangtoven (approved):

packages/wallet-sdk/src/relay/MobileRelay.ts
```diff
@@ -43,19 +43,49 @@ export class MobileRelay extends WalletLinkRelay {
...
         return;
       default:
+        navigatedToCBW = true;
         this.ui.openCoinbaseWalletDeeplink();
         break;
     }
+
+    // If the user navigated to the Coinbase Wallet app, then we need to check
+    // for unseen events once the user returns to the browser
+    if (navigatedToCBW) {
```
bangtoven, line 66: optional: could be defined as a function?

bangtoven (approved):
@vishnumad some optional comments

PR #981, opened 2023-08-15
feature: SDK ESM version

VladimirMikulic:
### _Summary_

In this PR, I added support for generating ESM version of the SDK.

Previously, the SDK was only distributed in CJS version which isn't optimized to be used in the browser and can hardly be tree shaken.

Here's the before (CJS) and after (ESM) bundle comparison of the SDK as a part of our app at work.

**Before: 142kb**

Screenshot_20230814_143919

**After: 121kb (15% decrease!)**

Screenshot_20230814_145443

It's also worth noting that this ESM migration opens the door to further SDK bundle optimizations which I plan to make in the future to reduce the overall size even more.

Compilation and tests succeeded. 
The SDK was also tested in our app at work.

Thanks!


VladimirMikulic:
@bangtoven could you please review the changes when you get a chance? Thanks!

bangtoven:
hey @VladimirMikulic sorry for the late response.
actually, we've been working on some major refactor of the lib, and we expect to reduce its size significantly by removing many of the dependencies. 
think we will revisit ESM support after finishing the refactor.
as this change now conflicts with the current master, i will close your pr for now, but feel free to open another one and let us know.
thanks for your contribution!

VladimirMikulic:
Appreciate your reply @bangtoven! Thanks for removing those dependencies. They are indeed the best way to reduce the bundle size.

PR #984, opened 2023-08-16
setDeepLinkConfig

bangtoven:

### DIFF
--- a/packages/wallet-sdk/src/provider/MobileRelayUI.ts
+++ b/packages/wallet-sdk/src/provider/MobileRelayUI.ts
@@ -55,7 +55,9 @@ export class MobileRelayUI implements WalletUI {
   }
 
   private redirectToCoinbaseWallet(walletLinkUrl?: string): void {
-    const baseUrl = this.useCustomScheme ? "cbwallet://" : "https://go.cb-w.com/";
+    const baseUrl = this.useCustomScheme
+      ? "cbwallet://"
+      : "https://go.cb-w.com/";
     const url = new URL("walletlink", baseUrl);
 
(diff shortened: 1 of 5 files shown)


packages/wallet-sdk/src/provider/MobileRelayUI.ts
```diff
@@ -41,12 +41,24 @@ export class MobileRelayUI implements WalletUI {
...
     this.useLocationMethod = useLocationMethod;
+    this.useCustomScheme = useCustomScheme;
   }
 
   private redirectToCoinbaseWallet(walletLinkUrl?: string): void {
-    const url = new URL("https://go.cb-w.com/walletlink");
+    const baseUrl = this.useCustomScheme
+      ? "cbwallet://"
+      : "https://go.cb-w.com/";
+    const url = new URL("walletlink", baseUrl);
```
vishnumad, line 61: The url opened seems to be `cbwallet:///walletlink?redirect_url...`. Note the three `///`

vishnumad:
I'm seeing some weird behavior when testing this change in a WKWebView. I updated your branch locally to use valid `cbwallet://walletlink` urls, but it seems like if I cancel the initial request and come back and send another request, it users the `https://go.cb-w.com` url again. 🤔 

image

packages/wallet-sdk/src/provider/MobileRelayUI.ts
bangtoven: thanks. what would be the optimal url for cbwallet walletlink path?
would it make the logic simpler if we have it the same format like `cbwallet://go.cb-w.com/walletlink`

bangtoven:
not gonna have this patch

PR #997, opened 2023-08-30
fix versioning gh aciton

bangtoven:

https://github.com/coinbase/coinbase-wallet-sdk/pull/999

### DIFF
--- a/.github/workflows/versioning.yml
+++ b/.github/workflows/versioning.yml
@@ -38,10 +38,10 @@ jobs:
         uses: actions/setup-node@v3
         with:
           node-version-file: '.nvmrc'
-          cache-dependency-path: 'yarn.lock'
+          cache-dependency-path: '**/yarn.lock'
 
       - name: Update version
-        run: npm version ${{ env.PACKAGE_VERSION }}
+        run: npm version ${{ env.PACKAGE_VERSION }} --no-git-tag-version --workspaces-update=false
 
       - name: Update version.ts file
         run: yarn prebuild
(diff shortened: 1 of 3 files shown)


.github/workflows/versioning.yml
```diff
@@ -41,8 +41,11 @@ jobs:
           cache-dependency-path: '**/yarn.lock'
 
       - name: Update version
-        run: yarn version --new-version ${{ env.PACKAGE_VERSION }} --no-git-tag-version
+        run: npm version ${{ env.PACKAGE_VERSION }} --no-git-tag-version --workspaces-update=false
```
bangtoven, line 44: `yarn version` is supported only on yarn 1.xx
as we bumped yarn version to 3.x, it no longer works.
tried multiple combinations of changes, but this seems to be the simplest solution.
for example, i tried `yarn plugin import version`, but it didn't allow to set the whole version value which this gh action is about. (we might migrate to this new method later)

.github/workflows/versioning.yml
```diff
@@ -41,8 +41,11 @@ jobs:
...
 
+      - name: Install NPM dependencies
+        run: yarn install --immutable
```
bangtoven, line 47: this is required to run any yarn command. applying the same changes made to main.yml gh action: https://github.com/coinbase/coinbase-wallet-sdk/commit/c7442551178e4a19254258796a8c791f6d6f8d11

packages/wallet-sdk/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@coinbase/wallet-sdk",
-  "version": "3.7.1",
+  "version": "3.9.0-canary.0",
```
bangtoven, line 3: verified: https://github.com/coinbase/coinbase-wallet-sdk/pull/999

cb-jake (approved):

jonesk7734 (commented):
Ok

PR #1000, opened 2023-08-31
fix npm publish

bangtoven:


packages/wallet-sdk/README.md
```diff
@@ -0,0 +1,126 @@
+# Coinbase Wallet SDK
```
bangtoven, line 1: npm publish requires reified README.md file. i tried symlink, but it didn't work.

cb-jake (approved):

PR #1004, opened 2023-09-12
Remove supportsAddressSwitching

fan-zhang-sv:
### _Summary_

- removed `supportsAddressSwitching` logic from sdk
- removed unused `isLedger` from sdk
- replaced duplicate code with existing func

Manually

### DIFF
--- a/packages/wallet-sdk/src/provider/CoinbaseWalletProvider.ts
+++ b/packages/wallet-sdk/src/provider/CoinbaseWalletProvider.ts
@@ -58,8 +58,6 @@ export interface CoinbaseWalletProviderOptions {
   relayProvider: () => Promise<WalletSDKRelayAbstract>;
   storage: ScopedLocalStorage;
   diagnosticLogger?: DiagnosticLogger;
-  supportsAddressSwitching?: boolean;
-  isLedger?: boolean;
 }
 
 interface AddEthereumChainParams {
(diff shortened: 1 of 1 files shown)


packages/wallet-sdk/src/provider/CoinbaseWalletProvider.ts
```diff
@@ -56,8 +56,6 @@ export interface CoinbaseWalletProviderOptions {
   relayProvider: () => Promise<WalletSDKRelayAbstract>;
   storage: ScopedLocalStorage;
   diagnosticLogger?: DiagnosticLogger;
-  supportsAddressSwitching?: boolean;
-  isLedger?: boolean;
```
fan-zhang-sv, line 60: isLedger was added from network switching project but actually never used in CBW. removing it now

packages/wallet-sdk/src/provider/CoinbaseWalletProvider.ts
```diff
@@ -179,7 +170,7 @@ export class CoinbaseWalletProvider
       },
     );
 
-    if (this._addresses.length > 0) {
+    if (this._isAuthorized()) {
```
fan-zhang-sv, line 173: theres an existing func to do this

packages/wallet-sdk/src/provider/CoinbaseWalletProvider.ts
```diff
@@ -646,7 +628,7 @@ export class CoinbaseWalletProvider
     return response;
   }
 
-  private _setAddresses(addresses: string[], isDisconnect?: boolean): void {
+  protected _setAddresses(addresses: string[], _?: boolean): void {
```
fan-zhang-sv, line 631: isDisconnect is only used for walletlinked connected extension. moving this to extension provider subclass

cb-jake (dismissed):

cb-jake (approved):

PR #1018, opened 2023-09-29
[WALL-25371] Remove RxJS from WalletLinkConnection

bangtoven:
### _Summary_

- removed RxJS and transitioned to TS promises while preserving the original data flow as closely as possible.
- please note that we prioritized maintaining consistency with the previous rxjs-based implementation. 
- we will revisit for a cleaner implementation with simpler logic.

manual + unit tests

https://github.com/coinbase/coinbase-wallet-sdk/assets/966662/927e5865-986b-4c42-9273-b0a0ef7acc01


packages/wallet-sdk/package.json
```diff
@@ -43,7 +43,6 @@
     "eventemitter3": "^5.0.1",
     "keccak": "^3.0.3",
     "preact": "^10.16.0",
-    "rxjs": "^6.6.3",
```
cb-jake, line 46: 😮 😲

packages/wallet-sdk/src/connection/RxWebSocket.test.ts
```diff
@@ -1,14 +1,13 @@
 import WS from 'jest-websocket-mock';
-import { Observable } from 'rxjs';
 
-import { ConnectionState, RxWebSocket } from './RxWebSocket';
+import { ConnectionState, WalletLinkWebSocket } from './RxWebSocket';
```
cb-jake, line 3: Should we rename this file name to WalletLinkWebSocket?

packages/wallet-sdk/src/connection/RxWebSocket.ts
```diff
@@ -132,7 +110,7 @@ export class RxWebSocket<T = object> {
     const { webSocket } = this;
     if (!webSocket) {
       this.pendingData.push(data);
-      this.connect().subscribe();
+      this.connect().then();
```
cb-jake, line 113: do we need the `.then()` can we just do `void this.connect();`

packages/wallet-sdk/src/connection/WalletLinkConnection.ts
```diff
@@ -101,111 +75,100 @@ export class WalletLinkConnection {
...
+            const connect = async () => {
+              // wait 5 seconds
+              await new Promise((resolve) => setTimeout(resolve, 5000));
+              // check whether it's destroyed again
+              if (!this.destroyed) {
+                try {
+                  // reconnect
+                  await ws.connect();
+                } catch {
+                  // retry()
```
cb-jake, line 102: what does retry do here and do we still need it?

packages/wallet-sdk/src/connection/WalletLinkConnection.ts
```diff
@@ -246,15 +223,14 @@ export class WalletLinkConnection {
     this.diagnostic?.log(EVENTS.STARTED_CONNECTING, {
       sessionIdHash: Session.hash(this.sessionId),
     });
-    this.ws.connect().subscribe();
+    this.ws.connect().then();
```
cb-jake, line 226: same question here

cb-jake:
🙌  Looks great! Will need to pull it down and test. I'll check it out on Monday!

packages/wallet-sdk/src/connection/RxWebSocket.test.ts
bangtoven: yes indeed. 
didn't make the change yet due to github diff issues. will address in a later pr.

packages/wallet-sdk/src/connection/WalletLinkConnection.ts
bangtoven: this was to note that it's replacing rxjs's retry operator. 
it can be removed. 👍

packages/wallet-sdk/src/connection/RxWebSocket.ts
bangtoven: this was to actually execute the promise that `this.connect()` returns, as `subscribe()` did.
i think one alternative could be `await this.connect()`, but didn't go with that since it would change the logic. this whole function used to just initiate the connection without waiting for it to complete and return.

bangtoven: actually you were right. i think i got confused while switching between rxjs and promises. 😭  updated.

packages/wallet-sdk/package.json
```diff
@@ -43,7 +43,6 @@
     "eventemitter3": "^5.0.1",
```
bangtoven, line 43: dear reviewers,
recommend to turn on `Hide whitespace`
https://github.com/coinbase/coinbase-wallet-sdk/pull/1018/files?diff=split&w=1
ty!

packages/wallet-sdk/src/connection/RxWebSocket.ts
```diff
@@ -1,27 +1,29 @@
 // Copyright (c) 2018-2023 Coinbase, Inc. <https://www.coinbase.com/>
 // Licensed under the Apache License, version 2.0
 
-import { BehaviorSubject, empty, Observable, of, Subject, throwError } from 'rxjs';
-import { flatMap, take } from 'rxjs/operators';
+import { ServerMessage } from './ServerMessage';
```
cb-jake, line 4: nit: should we rename this file to be `WebSocket` since we're no longer using RX?

bangtoven: https://github.com/coinbase/coinbase-wallet-sdk/pull/1018#discussion_r1346250503
@cb-jake yup. but then github considered it as a new file and made it hard to compare from the previous implementation. will rename the file as followup.

cb-jake: that makes sense

cb-jake (commented):
This looks good and testing seems to work. I think we should have someone with more rxjs experience should look at the PR before merging

vishnumad (approved):

packages/wallet-sdk/src/connection/WalletLinkConnection.ts
```diff
@@ -101,111 +73,98 @@ export class WalletLinkConnection {
...
+      // attempt to reconnect every 5 seconds when disconnected
+      this.diagnostic?.log(EVENTS.CONNECTED_STATE_CHANGE, {
+        state,
+        sessionIdHash: Session.hash(sessionId),
+      });
+
+      let connected = false;
+      switch (state) {
+        case ConnectionState.DISCONNECTED:
           // if DISCONNECTED and not destroyed
```
hieronymus777, line 89: Checking if it's intentional we don't skip the initial disconnect event.

packages/wallet-sdk/src/connection/WalletLinkConnection.ts
```diff
@@ -101,111 +73,98 @@ export class WalletLinkConnection {
...
+          break;
+
+        case ConnectionState.CONNECTED:
+          // perform authentication upon connection
+          try {
+            // if CONNECTED, authenticate, and then check link status
+            await this.authenticate();
+            this.sendIsLinked();
+            this.sendGetSessionConfig();
+            connected = true;
```
hieronymus777, line 113: Previously we would do 'distinctUntilChanged' - meaning we would only emit connected if it changes.  So we wouldn't emit two `true` events in a row.  Just making sure this doesn't have a side effect?

packages/wallet-sdk/src/connection/WalletLinkConnection.ts
```diff
@@ -101,111 +73,98 @@ export class WalletLinkConnection {
...
+                ws.connect().catch(() => {
+                  connect();
+                });
+              }
+            };
+            connect();
+          }
+          break;
+
+        case ConnectionState.CONNECTED:
```
hieronymus777, line 106: We were skipping 2 events before authenticating.  Not sure why this was here before but something to verify -are we sure we're not calling authenticate twice?

packages/wallet-sdk/src/connection/WalletLinkConnection.ts
```diff
@@ -246,15 +219,14 @@ export class WalletLinkConnection {
...
     });
-    this.ws.connect().subscribe();
+    this.ws.connect();
   }
 
   /**
    * Terminate connection, and mark as destroyed. To reconnect, create a new
    * instance of WalletSDKConnection
    */
   public destroy(): void {
```
hieronymus777, line 229: We were calling `unsubscribe` before which would clear the previous listeners.  I don't see that happening anymore.  Is there a possibility that the listeners will now get called after the connection is destroyed, for example through a race condition?

bangtoven: https://github.com/coinbase/coinbase-wallet-sdk/pull/1040

packages/wallet-sdk/src/connection/WalletLinkConnection.ts
bangtoven: for the record:
with rxjs subject, it used to emit a default value `DISCONNECTED` whenever the observable was subscribed
by replacing it with callback listener, we expect `connectionStateListener` only gets invoked when there’s actual state changes. so we don’t need `skip(1)` for `DISCONNECTED`

packages/wallet-sdk/src/connection/WalletLinkConnection.ts
bangtoven: for `ConnectionState.CONNECTED` check which used to have `skip(2)` , i believe it was to ignore initial `DISCONNECTED` and `CONNECTING` states as the comment said. we don’t need this skipping anymore for this code block as we moved to simple switch syntax instead of individual rx subscription.

PR #1027, opened 2023-10-11
Small Refactor

fan-zhang-sv:
### _Summary_
- removed connected and chainId from WalletUIOptions interface, and moved it 1 level down to LinkFlow
- changed publishEvent to protected

Manually

### DIFF
--- a/packages/wallet-sdk/src/provider/WalletLinkRelayUI.ts
+++ b/packages/wallet-sdk/src/provider/WalletLinkRelayUI.ts
@@ -34,8 +34,6 @@ export class WalletLinkRelayUI implements WalletUI {
       sessionId: options.session.id,
       sessionSecret: options.session.secret,
       linkAPIUrl: options.linkAPIUrl,
-      connected: false,
-      chainId: 1,
       isParentConnection: false,
     });
   }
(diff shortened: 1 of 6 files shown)


packages/wallet-sdk/src/provider/WalletLinkRelayUI.ts
```diff
@@ -34,8 +34,8 @@ export class WalletLinkRelayUI implements WalletUI {
       sessionId: options.session.id,
       sessionSecret: options.session.secret,
       linkAPIUrl: options.linkAPIUrl,
-      connected: options.connected,
```
bangtoven, line 37: can we just remove these as well?

fan-zhang-sv: ahhhh right, updated!

packages/wallet-sdk/src/relay/WalletLinkRelay.ts
```diff
@@ -302,8 +302,6 @@ export class WalletLinkRelay extends WalletSDKRelayAbstract {
       version: this.options.version,
       darkMode: this.options.darkMode,
       session,
-      connected: false,
-      chainId: this.dappDefaultChain,
```
bangtoven, line 306: not sure how this is being used 🤔

bangtoven (approved):

packages/wallet-sdk/src/relay/WalletLinkRelay.ts
fan-zhang-sv: these 2 are for the QR code dialog. and chainId is for when dapp provides a preferred default network that is not Ethereum

fan-zhang-sv: Tested with test-dapp, and RN, RN correctly parsed the chainId and connected via qr code w/o issue.

PR #1038, opened 2023-10-26
Add METHOD_NOT_IMPLEMENTED diagnostic event

bangtoven:
### _Summary_

https://coinbase.slack.com/archives/C03QN57NMHV/p1698264813994269

### DIFF
--- a/packages/wallet-sdk/src/provider/CoinbaseWalletProvider.ts
+++ b/packages/wallet-sdk/src/provider/CoinbaseWalletProvider.ts
@@ -804,6 +804,11 @@ export class CoinbaseWalletProvider extends EventEmitter implements Web3Provider
         return this._wallet_watchAsset(params);
     }
 
+    this.diagnostic?.log(EVENTS.METHOD_NOT_IMPLEMENTED, {
+      method: request.method,
+      sessionIdHash: this._relay ? Session.hash(this._relay.session.id) : undefined,
+    });
+
     const relay = await this.initializeRelay();
     return relay.makeEthereumJSONRPCRequest(request, this.jsonRpcUrl);
   }
--- a/packages/wallet-sdk/src/connection/DiagnosticLogger.ts
+++ b/packages/wallet-sdk/src/connection/DiagnosticLogger.ts
@@ -46,5 +46,6 @@ export const EVENTS = {
   WEB3_REQUEST: 'walletlink_sdk.web3.request',
   WEB3_REQUEST_PUBLISHED: 'walletlink_sdk.web3.request_published',
   WEB3_RESPONSE: 'walletlink_sdk.web3.response',
+  METHOD_NOT_IMPLEMENTED: 'walletlink_sdk.method_not_implemented',
   UNKNOWN_ADDRESS_ENCOUNTERED: 'walletlink_sdk.unknown_address_encountered',
 };


packages/wallet-sdk/src/provider/CoinbaseWalletProvider.ts
```diff
@@ -804,6 +804,11 @@ export class CoinbaseWalletProvider extends EventEmitter implements Web3Provider
         return this._wallet_watchAsset(params);
     }
 
+    this.diagnostic?.log(EVENTS.METHOD_NOT_IMPLEMENTED, {
+      method: request.method,
+      sessionIdHash: this._relay ? Session.hash(this._relay.session.id) : undefined,
```
bangtoven, line 809: following the existing pattern 
image

cb-jake (approved):

packages/wallet-sdk/src/provider/CoinbaseWalletProvider.ts
```diff
@@ -804,6 +804,11 @@ export class CoinbaseWalletProvider extends EventEmitter implements Web3Provider
         return this._wallet_watchAsset(params);
     }
 
+    this.diagnostic?.log(EVENTS.METHOD_NOT_IMPLEMENTED, {
```
hieronymus777, line 807: This will emit a `METHOD_NOT_IMPLEMENTED` event for all json rpc calls (i.e. if the dapp is using the web3 provider to call `eth_getBalance` and so on).  I don't think that's correct since those methods are implemented - by the json rpc endpoint.

I think what we want is to catch the response from `makeEthereumJSONRPCRequest` and if it's an unimplemented method error, then emit the event. 

Something like
```
    return relay.makeEthereumJSONRPCRequest(request, this.jsonRpcUrl).catch((err) => {
      if (err.code === standardErrorCodes.rpc.methodNotFound) {
        this.diagnostic?.log(EVENTS.METHOD_NOT_IMPLEMENTED, {
          method: request.method,
          sessionIdHash: this._relay ? Session.hash(this._relay.session.id) : undefined,
        });
      }

      throw err;
    });
```

hieronymus777: Or to be extra safe
```
    return relay.makeEthereumJSONRPCRequest(request, this.jsonRpcUrl).catch((err) => {
      if (err.code === standardErrorCodes.rpc.methodNotFound || err.code === standardErrorCodes.rpc.methodNotSupported) {
        this.diagnostic?.log(EVENTS.METHOD_NOT_IMPLEMENTED, {
          method: request.method,
          sessionIdHash: this._relay ? Session.hash(this._relay.session.id) : undefined,
        });
      }

      throw err;
    });
```

cb-jake: Good catch. I should have noticed this.

bangtoven: @hieronymus777 thanks for pointing this out. i totally missed.
will update soon. cc @cb-jake

bangtoven: https://github.com/coinbase/coinbase-wallet-sdk/pull/1040

PR #1045, opened 2023-10-31
CoinbaseWalletProvider should NOT listen to 'addressChanged' event

gaurav-cb:
### _Summary_

CoinbaseWalletProvider should NOT listen to 'addressChanged' event 

Unit tests

### DIFF
--- a/packages/wallet-sdk/src/provider/CoinbaseWalletProvider.ts
+++ b/packages/wallet-sdk/src/provider/CoinbaseWalletProvider.ts
@@ -186,10 +186,6 @@ export class CoinbaseWalletProvider extends EventEmitter implements Web3Provider
         const jsonRpcUrl = event.data.data.jsonRpcUrl ?? this.jsonRpcUrl;
         this.updateProviderInfo(jsonRpcUrl, Number(_chainId));
       }
-
-      if (event.data.data.action === 'addressesChanged') {
-        this._setAddresses(event.data.data.addresses);
-      }
     });
   }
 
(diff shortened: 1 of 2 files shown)


packages/wallet-sdk/src/provider/CoinbaseWalletProvider.ts
```diff
@@ -187,8 +187,8 @@ export class CoinbaseWalletProvider extends EventEmitter implements Web3Provider
         this.updateProviderInfo(jsonRpcUrl, Number(_chainId));
       }
 
-      if (event.data.data.action === 'addressChanged') {
```
cb-jake, line 190: Would we want to maintain both for backwards compat?

gaurav-cb: older dapps that haven't upgraded to the new SDK would be fine because we are still emitting the 'addressChanged' event from extension.

So bascially:
- extension sends both 'addressChanged' and 'addressedChanged'
- if on old SDK, listen to 'addressChanged'
- if on new SDK, listen to 'addressesChanged'

cb-jake: One issue though, we will probably release the SDK before the extension with those changes is rolled out 100% to prod

cb-jake:
What about walletlink mobile? Does this change impact the mobile connection

packages/wallet-sdk/src/provider/CoinbaseWalletProvider.ts
gaurav-cb: We can wait on this PR being release

bangtoven (approved):

PR #1059, opened 2023-11-11
Refactor session config handling and cryptography logic for improved abstraction

bangtoven:
### _Summary_

https://coinbase.slack.com/archives/C03QN57NMHV/p1699906732036009
- consolidate listener functions 
- migrate session config handling from `Relay` to `Connection`
- introduce WalletLinkConnectionCipher to encapsulate secret token management and cryptography logic
- migrate cipher logic from `Relay` to `Connection`
- add / update tests

### _How did you test your changes?_

new & existing tests

PR #1063, opened 2023-11-14
WalletLinkWebSocketUpdateListener

bangtoven:
### _Summary_

- Apply the "listener" pattern introduced in https://github.com/coinbase/coinbase-wallet-sdk/pull/1059


packages/wallet-sdk/src/connection/WalletLinkWebSocket.ts
```diff
@@ -1,30 +1,48 @@
...
+  ServerMessage,
+  ServerMessageFail,
+  ServerMessageOK,
+} from './ServerMessage';
 
-export enum ConnectionState {
-  DISCONNECTED,
-  CONNECTING,
-  CONNECTED,
-}
```
bangtoven, line 10: arbitrary client-side-only concept

PR #1065, opened 2023-11-15
Adding More Methods for Testdapp

fan-zhang-sv:

https://6555636792e30f11237feba4--spontaneous-alpaca-d199e2.netlify.app/


apps/testapp/src/components/Layout.tsx
```diff
@@ -1,26 +1,59 @@
...
+} from '@chakra-ui/react';
 
-import { useCBWSDK } from '../context/CBWSDKProvider';
+import { sdkVersions, useCBWSDK } from '../context/CBWSDKProvider';
 
 type LayoutProps = {
   children: React.ReactNode;
 };
 
+export const WIDTH_2XL = '1536px';
```
fan-zhang-sv, line 20: 2XL from taiwind

package.json
```diff
@@ -27,5 +27,8 @@
     "eslint-plugin-unused-imports": "^3.0.0",
     "prettier": "^3.0.0"
   },
-  "packageManager": "yarn@3.6.1"
+  "packageManager": "yarn@3.6.1",
+  "dependencies": {
+    "@metamask/eth-sig-util": "^7.0.0"
```
fan-zhang-sv, line 32: for verifying signature

apps/testapp/src/components/RpcMethods/RpcMethod.ts
```diff
@@ -0,0 +1,14 @@
...
+  data: Record<string, string>;
+};
+
+export type RpcMethod = {
+  connected?: boolean;
+  method: string;
+  params: Array<{ key: string; required?: boolean }>;
+  format?: (data: Record<string, string>) => FormattedParamsType[];
+  shortcuts?: ShortcutType[];
+};
```
bangtoven, line 14: @fan-zhang-sv  we already discussed offline, but for the record here.
let's explore whether we can achieve this without introducing new type.

`RpcMethod` could be misleading. it seems this type is an extension of `Request` type. how about something like
```suggestion
export type RpcRequestInput = {
  method: RPCMethod; // strongly typed
};
```
`ShortcutType` might better be managed separately from the RPC method, probably a map between `method` and set of `shortcuts`? it's just rough idea.

i will also take a look at the SDK code to see if any of the internal types we have could be exposed in some way for type checking.

fan-zhang-sv: yup yup,

- added separate ShortcutType
- renamed to RpcRequestInput, and have method field strongly typed
- created mapping to load shortcuts for method

packages/wallet-sdk/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@coinbase/wallet-sdk",
-  "version": "3.9.0-canary.4",
+  "version": "3.9.0-canary.5",
```
fan-zhang-sv, line 3: bump version up, as we have canary5 published already

bangtoven (approved):
🔥

bangtoven (approved):

PR #1079, opened 2023-11-22
Misc update

fan-zhang-sv:

### DIFF
--- a/apps/testapp/src/context/CBWSDKProvider.tsx
+++ b/apps/testapp/src/context/CBWSDKProvider.tsx
@@ -8,6 +8,7 @@ type CBWSDKProviderProps = {
 };
 
 const CBWSDKContext = React.createContext(null);
+const SELECTED_SDK_KEY = 'selected_sdk_version';
 
 export const sdkVersions = ['master', '3.9', '3.7'] as const;
 export type SDKVersionType = (typeof sdkVersions)[number];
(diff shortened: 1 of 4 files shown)


apps/testapp/src/context/CBWSDKProvider.tsx
```diff
@@ -27,27 +28,41 @@ const dynamicallyImportSDK = (version: SDKVersionType) => {
...
+      const cbwprovider = cbwsdk.makeWeb3Provider('http');
+      setProvider(cbwprovider);
+    }
   }, [version]);
 
+  const setSDKVersion = (version: SDKVersionType) => {
+    localStorage.setItem(SELECTED_SDK_KEY, version);
+    setVersion(version);
+  };
+
```
fan-zhang-sv, line 59: save selected sdkversion in localstorage, so sdk remians the same version on each refresh

packages/wallet-sdk/src/relay/walletlink/type/Web3Method.ts
```diff
@@ -1,18 +1,21 @@
...
+  'scanQRCode',
+  'generic',
+  'childRequestEthereumAccounts',
+  'addEthereumChain',
+  'switchEthereumChain',
+  'makeEthereumJSONRPCRequest',
+  'watchAsset',
+  'selectProvider',
+  'connectAndSignIn',
+] as const;
```
fan-zhang-sv, line 19: we can derive union type from array, but not other way around. doing this so wallet can check is a string is valid web3method

packages/wallet-sdk/src/relay/walletlink/type/Web3Request.ts
```diff
@@ -23,6 +23,9 @@ type _Web3Request =
         appLogoUrl: string | null;
       };
     }
+  | {
+      method: 'childRequestEthereumAccounts';
```
fan-zhang-sv, line 27: another type that is used between WL extension and RN, adding it here as the new `WalletLinkEventData` tightened the types for walletlinkConnection

packages/wallet-sdk/src/relay/walletlink/type/Web3Method.ts
bangtoven: love it

bangtoven (approved):

bangtoven (approved):

bangtoven (approved):

PR #1121, opened 2024-02-24
Fix/deeplink from webview for android

pedrorosarioo:
### _Summary_
The href replacing approach has fixed the deeplink issue for WebView on IOS, but caused a regression on Android. The anchor tag approach fixes it for both platforms

Manually tested


willn-cb (approved):
Tested locally and confirmed this works for the following cases for Android and iOS:
- Mobile app with SDK loaded in a WebView
- Mobile Safari
- Mobile Chrome

Mandog1120 (commented):
c57516aa352a0a5ceecbff264e03a5b29bd4038b

PR #1134, opened 2024-03-15
initialize SCWStateManager with dapp chainIds

fan-zhang-sv:
### _Summary_

We follow the app's `chainIds` param to set the default network for SDK, but we missed out SCWStateManager, so it's always sending requests on mainnet. This PR updated SCWStateManager to take `chainIds` for fallback

Manually

### DIFF
--- a/packages/wallet-sdk/src/sign/scw/SCWStateManager.ts
+++ b/packages/wallet-sdk/src/sign/scw/SCWStateManager.ts
@@ -41,7 +41,7 @@ export class SCWStateManager {
     }
 
     this._accounts = accounts || [];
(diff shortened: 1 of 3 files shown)


packages/wallet-sdk/src/sign/scw/SCWStateManager.ts
```diff
@@ -41,7 +41,7 @@ export class SCWStateManager {
     }
```
nateReiners, line 41: nit would place updateListeners after _activeChain and _accounts are set

nateReiners (approved):

PR #1138, opened 2024-03-21
Add new wallet RPCs 

nateReiners:
### _Summary_
Adds support for 3 new RPC methods proposed here: https://hackmd.io/@lsr/proposal-for-new-wallet-rpcs 
Note: `wallet_showTransactionStatus` is not being supported for now but may be added at a later date.

- `wallet_getCapabilities`
  - RPC for an application to receive information about the features that a connected wallet supports.
 Comment
- `wallet_sendTransaction`
  - Requests that a wallet submits a transaction. Sender and chain id are top level because all calls should be sent from the same sender on the same chain. The items in the calls field are only those that are shared by all transaction types. Any other fields that a wallet may need to submit a transaction should be handled by the wallet.
- `wallet_getTransactionStatus`
  - Inspired by EIP-5792’s wallet_getBundleStatus. For each submitted call to wallet_sendTransaction, returns a status of CONFIRMED or PENDING and a subset of fields that eth_getTransactionReceipt returns, excluding any fields that may differ across wallet implementations.

locally - will wait for backend work for `wallet_getTransactionStatus` to be completed and verify before merging

### DIFF
--- a/apps/testapp/src/components/RpcMethods/method/walletTxMethods.ts
+++ b/apps/testapp/src/components/RpcMethods/method/walletTxMethods.ts
@@ -1,0 +1,44 @@
+import { parseMessage } from '../shortcut/ShortcutType';
+import { RpcRequestInput } from './RpcRequestInput';
+
+const walletGetCapabilities: RpcRequestInput = {
+  method: 'wallet_getCapabilities',
+  params: [],
+};
+
+const walletSendTransaction: RpcRequestInput = {
+  method: 'wallet_sendTransaction',
+  params: [
+    { key: 'chainId', required: true },
+    { key: 'sender', required: true },
+    { key: 'calls', required: true },
+    { key: 'capabilities', required: true },
+  ],
+  format: (data: Record<string, string>) => [
+    {
+      chainId: data.chainId,
+      sender: data.sender,
+      calls: data.calls,
+      capabilities: parseMessage(data.capabilities),
+    },
+  ],
+};
+
+const walletGetTransactionStatus: RpcRequestInput = {
+  method: 'wallet_getTransactionStatus',
+  params: [{ key: 'params', required: true }],
+  format: (data: Record<string, string>) => [data.params],
+};
+
+const walletShowTransactionStatus: RpcRequestInput = {
+  method: 'wallet_showTransactionStatus',
+  params: [{ key: 'params', required: true }],
+  format: (data: Record<string, string>) => [data.params],
+};
+
+export const walletTxMethods = [
+  walletGetCapabilities,
+  walletGetTransactionStatus,
+  walletShowTransactionStatus,
+  walletSendTransaction,
+];
--- a/apps/testapp/src/components/RpcMethods/method/walletTxMethods.ts
+++ b/apps/testapp/src/components/RpcMethods/method/walletTxMethods.ts
@@ -34,15 +34,8 @@ const walletGetTransactionStatus: RpcRequestInput = {
   format: (data: Record<string, string>) => [data.params],
 };
 
-const walletShowTransactionStatus: RpcRequestInput = {
-  method: 'wallet_showTransactionStatus',
-  params: [{ key: 'params', required: true }],
-  format: (data: Record<string, string>) => [data.params],
-};
-
 export const walletTxMethods = [
   walletGetCapabilities,
   walletGetTransactionStatus,
-  walletShowTransactionStatus,
   walletSendTransaction,
 ];
--- a/apps/testapp/src/components/RpcMethods/method/walletTxMethods.ts
+++ b/apps/testapp/src/components/RpcMethods/method/walletTxMethods.ts
@@ -12,13 +12,17 @@ const walletSendTransaction: RpcRequestInput = {
     { key: 'chainId', required: true },
     { key: 'sender', required: true },
     { key: 'calls', required: true },
-    { key: 'capabilities', required: true },
+    { key: 'version', required: true },
+    { key: 'gas', required: false },
+    { key: 'capabilities', required: false },
   ],
   format: (data: Record<string, string>) => [
     {
       chainId: data.chainId,
       sender: data.sender,
       calls: data.calls,
+      version: data.version,
+      gas: data.gas,
       capabilities: parseMessage(data.capabilities),
     },
   ],
--- a/apps/testapp/src/components/RpcMethods/shortcut/walletTxShortcuts.ts
+++ b/apps/testapp/src/components/RpcMethods/shortcut/walletTxShortcuts.ts
@@ -1,0 +1,22 @@
+import { ADDR_TO_FILL } from './const';
+import { ShortcutType } from './ShortcutType';
+
+const walletSendTransactionShortcuts: ShortcutType[] = [
+  {
+    key: 'wallet_sendTransaction',
+    data: {
+      chainId: '84532',
+      sender: ADDR_TO_FILL,
+      calls: [],
+      capabilities: {
+        paymaster: {
+          url: 'https://paymaster.base.org',
+        },
+      },
+    },
+  },
+];
+
+export const walletTxShortcutsMap = {
+  wallet_sendTransaction: walletSendTransactionShortcuts,
+};
--- a/apps/testapp/src/components/RpcMethods/shortcut/walletTxShortcuts.ts
+++ b/apps/testapp/src/components/RpcMethods/shortcut/walletTxShortcuts.ts
@@ -8,6 +8,7 @@ const walletSendTransactionShortcuts: ShortcutType[] = [
       chainId: '84532',
       sender: ADDR_TO_FILL,
       calls: [],
+      version: '1',
       capabilities: {
         paymaster: {
           url: 'https://paymaster.base.org',
--- a/packages/wallet-sdk/src/sign/SignRequestHandler.ts
+++ b/packages/wallet-sdk/src/sign/SignRequestHandler.ts
@@ -108,7 +108,6 @@ export class SignRequestHandler implements RequestHandler {
       'wallet_getCapabilities',
       'wallet_sendTransaction',
       'wallet_getTransactionStatus',
-      'wallet_showTransactionStatus',
     ];
 
     return methodsThatRequireSigning.includes(request.method);
--- a/packages/wallet-sdk/src/sign/SignRequestHandler.ts
+++ b/packages/wallet-sdk/src/sign/SignRequestHandler.ts
@@ -105,6 +105,10 @@ export class SignRequestHandler implements RequestHandler {
       'wallet_addEthereumChain',
       'wallet_switchEthereumChain',
       'wallet_watchAsset',
+      'wallet_getCapabilities',
+      'wallet_sendTransaction',
+      'wallet_getTransactionStatus',
+      'wallet_showTransactionStatus',
     ];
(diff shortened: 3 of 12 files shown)


apps/testapp/src/components/RpcMethods/method/walletTxMethods.ts
```diff
@@ -0,0 +1,44 @@
...
+  params: [],
+};
+
+const walletSendTransaction: RpcRequestInput = {
+  method: 'wallet_sendTransaction',
+  params: [
+    { key: 'chainId', required: true },
+    { key: 'sender', required: true },
+    { key: 'calls', required: true },
+    { key: 'capabilities', required: true },
```
lukasrosario, line 15: capabilities are optional. also missing here are `gas` and `version`. `gas` is optional and `version` is required

apps/testapp/src/components/RpcMethods/shortcut/walletTxShortcuts.ts
```diff
@@ -0,0 +1,22 @@
...
+const walletSendTransactionShortcuts: ShortcutType[] = [
+  {
+    key: 'wallet_sendTransaction',
+    data: {
+      chainId: '84532',
+      sender: ADDR_TO_FILL,
+      calls: [],
+      capabilities: {
+        paymaster: {
+          url: 'https://paymaster.base.org',
```
lukasrosario, line 13: should update to include version as specified above

apps/testapp/src/components/RpcMethods/method/walletTxMethods.ts
nateReiners: thanks, will update

apps/testapp/src/components/RpcMethods/shortcut/walletTxShortcuts.ts
nateReiners: thanks! will update

packages/wallet-sdk/src/sign/SignRequestHandler.ts
```diff
@@ -105,6 +105,10 @@ export class SignRequestHandler implements RequestHandler {
       'wallet_addEthereumChain',
       'wallet_switchEthereumChain',
       'wallet_watchAsset',
+      'wallet_getCapabilities',
+      'wallet_sendTransaction',
+      'wallet_getTransactionStatus',
+      'wallet_showTransactionStatus',
```
lukasrosario, line 111: think we've landed on cutting wallet_showTransactionStatus. seems like it would be easy to add back if we need it but thinking we should remove here for now. lmk if im wrong though, if it would be too much of a headache to add back later then we can just leave it as is

packages/wallet-sdk/src/sign/scw/message/type/Action.ts
```diff
@@ -127,6 +131,36 @@ export type AddEthereumChainAction = {
...
+export type WalletSendTransactionAction = {
+  method: SupportedEthereumMethods.WalletSendTransaction;
+  params: {
+    chainId: number;
+    sender: string;
+    calls: {
+      target: string;
+      value: string;
+      data: string;
+      gas: string;
```
lukasrosario, line 148: gas is optional and top level (with chainId and sender). and need a version string, top level too.

packages/wallet-sdk/src/sign/scw/message/type/Action.ts
```diff
@@ -127,6 +131,36 @@ export type AddEthereumChainAction = {
   };
 };
 
+export type GetCapacitiesAction = {
+  method: SupportedEthereumMethods.WalletGetCapacities;
+  params: null;
+};
```
lukasrosario, line 137: should be called capacities --> capabilities. also takes a single argument: a chain id. capabilities could differ per chain so method will take a chain id and return capabilities for that chain id. also think it should be hex, chain ids are uint256

packages/wallet-sdk/src/sign/scw/message/type/Action.ts
```diff
@@ -127,6 +131,36 @@ export type AddEthereumChainAction = {
...
+
+export type WalletSendTransactionAction = {
+  method: SupportedEthereumMethods.WalletSendTransaction;
+  params: {
+    chainId: number;
```
lukasrosario, line 142: think this should be hex

packages/wallet-sdk/src/sign/scw/message/type/Action.ts
nateReiners: I think this would mean that for every chain switch, the popup would need to open briefly :/ . Not ideal.

apps/testapp/src/components/RpcMethods/method/walletTxMethods.ts
nateReiners: addressed in [6178c33](https://github.com/coinbase/coinbase-wallet-sdk/pull/1138/commits/6178c33e20166ef8e819447e8ed74dbd1299dfb7)

apps/testapp/src/components/RpcMethods/shortcut/walletTxShortcuts.ts
nateReiners: addressed in [6178c33](https://github.com/coinbase/coinbase-wallet-sdk/pull/1138/commits/6178c33e20166ef8e819447e8ed74dbd1299dfb7)

packages/wallet-sdk/src/sign/SignRequestHandler.ts
nateReiners: Sounds good, if we add it back dapps will have to update their code anyway to start using it. The only difference is just adding a version # bump to those codechanges

addressed in [35edf8f](https://github.com/coinbase/coinbase-wallet-sdk/pull/1138/commits/35edf8fa992029ab3feabc878b9891d7a7b0036f)

packages/wallet-sdk/src/sign/scw/message/type/Action.ts
nateReiners: thank you, addressed in [cd1eb13](https://github.com/coinbase/coinbase-wallet-sdk/pull/1138/commits/cd1eb1360fb42b48a38b88c0bc59db3c2c3f9b1f)

packages/wallet-sdk/src/sign/scw/message/type/Action.ts
nateReiners: thanks, addressed in [4c4ccb7](https://github.com/coinbase/coinbase-wallet-sdk/pull/1138/commits/4c4ccb74d650bdcbbe02ef34aba545d8ce449104)

packages/wallet-sdk/src/sign/scw/message/type/Action.ts
nateReiners: ok per discussion elsewhere we'll receive and store capabilities per chain during the handshake - updates here [2e350be](https://github.com/coinbase/coinbase-wallet-sdk/pull/1138/commits/2e350befe010b3902d7353498a2b5b4baa6e6f13)

fan-zhang-sv (commented):
nice! left some comments

packages/wallet-sdk/src/sign/scw/SCWStateManager.ts
```diff
@@ -41,7 +47,7 @@ export class SCWStateManager {
     }
 
     this._accounts = accounts || [];
-    this._activeChain = chain || { id: options.appChainIds?.[0] ?? 1 };
+    this._activeChain = chain || { id: 1 };
```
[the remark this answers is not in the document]
nateReiners: good catch! addressed here [db84975](https://github.com/coinbase/coinbase-wallet-sdk/pull/1138/commits/db8497543f23f37eb8c22501b9a0793c35ed49cb)

packages/wallet-sdk/src/sign/walletlink/WLSigner.ts
```diff
@@ -27,6 +28,10 @@ export class WLSigner implements Signer {
   }
 
   async request<T>(requestArgs: RequestArguments): Promise<T> {
+    if (requestArgs.method === 'wallet_getCapabilities') {
+      // TODO: remove this check; pass this request to WLRelayAdapter
+      Promise.reject(standardErrors.rpc.methodNotSupported());
+    }
```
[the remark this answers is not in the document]
nateReiners: agreed! addressed here, [4905e7e](https://github.com/coinbase/coinbase-wallet-sdk/pull/1138/commits/4905e7e6069e666d25738631b2c08e76fd38a395)

packages/wallet-sdk/src/sign/scw/SCWSigner.ts
```diff
@@ -119,6 +129,39 @@ export class SCWSigner implements Signer {
         // https://eips.ethereum.org/EIPS/eip-3326#wallet_switchethereumchain
         return switched ? (null as T) : undefined;
       }
+      case SupportedEthereumMethods.WalletGetCapabilities: {
+        const params = request.params as GetCapabilitiesAction['params'];
+        const walletCapabilities = this.stateManager.walletCapabilities?.[params.chainId];
```
lukasrosario, line 134: we decided this should just return all chain ids capabilities. ie it does not take a chain id param, and returns an object where the keys are chain ids and values are capabilities for that chain id

packages/wallet-sdk/src/sign/walletlink/WLSigner.ts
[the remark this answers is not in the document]
fan-zhang-sv: hey sorry! i was wrong about this
 just realized
1. blocking it from SDK would be bad for enabling this in the furture
2. seems like WL BE is already blocking unrecognized methods, even if we dont do anything

think we can remove what we doing here at WLSigner, thx!

image

packages/wallet-sdk/src/sign/scw/SCWSigner.ts
nateReiners: Ah okay makes sense so dapps can just check the capabilities once and cache the result, thanks for clarifying

packages/wallet-sdk/src/sign/walletlink/WLSigner.ts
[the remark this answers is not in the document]
nateReiners: Ok then we can allow them all including `wallet_getTransactionStatus`

packages/wallet-sdk/src/sign/scw/SCWSigner.ts
nateReiners: addressed in [f7bae2d](https://github.com/coinbase/coinbase-wallet-sdk/pull/1138/commits/f7bae2dfaa24b3d9e72c7d2ef144388b998914ed) and [da15e6c](https://github.com/coinbase/coinbase-wallet-sdk/pull/1138/commits/da15e6cdef101533d559479e8c55ea645ce0dbf0)

packages/wallet-sdk/src/sign/walletlink/WLSigner.ts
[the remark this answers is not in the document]
nateReiners: makes sense, addressed in [26665de](https://github.com/coinbase/coinbase-wallet-sdk/pull/1138/commits/26665dee8d8564acc074b0f4e5c588290b6ea97c)

PR #1143, opened 2024-03-28
default chain logic in testdapp

fan-zhang-sv:
### _Summary_

Switch to use the new init config to set default chain in the testdapp


nateReiners (approved):
thanks!

PR #1154, opened 2024-04-03
fix bad message formatting

nateReiners:


fan-zhang-sv (commented):
@nateReiners  hmm were we having any issue with the old format function? i tested locally against prod SCW, without this, the example message wouldn't work tho

PR #1163, opened 2024-04-09
SDK Pass Version on Handshake + Error Handling for Signer Selection

fan-zhang-sv:
### _Summary_

1. SDK Pass Version on Handshake
2. When user close popup before selecting the signer, SDK can now correctly handling it and throw back to dapp


nateReiners (approved):
nice

PR #1164, opened 2024-04-09
Extension Detection

fan-zhang-sv:
### _Summary_

- only return injected provider, when coinbaseWalletExtension is injected and coinbaseWalletExtensionSigner is not
- when FE returns selected signer as extension, grab the injected coinbaseWalletExtensionSigner
- sends coinbaseWalletExtensionSigner status to FE for selection

When extensionSigner is injected, always open SCW popup and show extension option
https://github.com/coinbase/coinbase-wallet-sdk/assets/22125939/16e27720-6e8d-4a78-b698-36305e6e04af

when extensionSigner is not injected, but old extension provider is injected
https://github.com/coinbase/coinbase-wallet-sdk/assets/22125939/c0050356-4afa-409e-a016-cb0fe81c0be6

When error grabbing injected extensionSigner, reset everything
https://github.com/coinbase/coinbase-wallet-sdk/assets/22125939/455d5fd6-afb0-4579-83f0-b4c1cb781fdd

### DIFF
--- a/apps/testapp/src/context/CBWSDKReactContextProvider.tsx
+++ b/apps/testapp/src/context/CBWSDKReactContextProvider.tsx
@@ -30,10 +30,12 @@ declare global {
 
 if (typeof window !== 'undefined') {
   window.setPopupUrl = (url: string) => {
-    const handler = (window.ethereum as any).handlers.find(
+    const handler = (window.ethereum as any).handlers?.find(
       (h: any) => h instanceof SignRequestHandler
     );
-    handler.popupCommunicator.url = new URL(url);
+    if (handler && handler.popupCommunicator) {
+      handler.popupCommunicator.url = new URL(url);
+    }
   };
 }
 
--- a/packages/wallet-sdk/src/index.ts
+++ b/packages/wallet-sdk/src/index.ts
@@ -3,6 +3,7 @@
 import { CoinbaseWalletProvider } from './CoinbaseWalletProvider';
 import { CoinbaseWalletSDK } from './CoinbaseWalletSDK';
 import { ProviderInterface } from './core/type/ProviderInterface';
+import { Signer } from './sign/SignerInterface';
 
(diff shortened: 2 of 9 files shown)


apps/testapp/src/context/CBWSDKReactContextProvider.tsx
```diff
@@ -30,10 +30,12 @@ declare global {
...
 if (typeof window !== 'undefined') {
   window.setPopupUrl = (url: string) => {
-    const handler = (window.ethereum as any).handlers.find(
+    const handler = (window.ethereum as any).handlers?.find(
       (h: any) => h instanceof SignRequestHandler
     );
-    handler.popupCommunicator.url = new URL(url);
+    if (handler && handler.popupCommunicator) {
+      handler.popupCommunicator.url = new URL(url);
+    }
```
fan-zhang-sv, line 38: a bug in playground, exception when extension is installed

packages/wallet-sdk/src/index.ts
```diff
@@ -13,6 +14,7 @@ declare global {
     CoinbaseWalletProvider: typeof CoinbaseWalletProvider;
     ethereum?: ProviderInterface;
     coinbaseWalletExtension?: ProviderInterface;
+    coinbaseWalletExtensionSigner?: Signer;
```
fan-zhang-sv, line 17: my reason to use `coinbaseWalletExtensionSigner` instead of what we talked before, "coinbaseWalletInjectedSigner", is i think it's always better to have more granularity. cc: @bangtoven

nateReiners: makes sense to me

nateReiners (approved):

packages/wallet-sdk/src/index.ts
Schmeltzermark: That makes sense to me also.

PR #1165, opened 2024-04-09
[TRIVIAL] add data field to eth_sendTransaction in test app

spencerstock:
### _Summary_

Adding a data field so that we can pass arbitrary call data into the playground test app


nateReiners (approved):
thanks

PR #1166, opened 2024-04-10
disconnect cleanup

nateReiners:
### _Summary_
- remove `sdk.disconnect()` method
- add disconnect to `ProviderInterface` otherwise TS complains if anyone tries to call provider.disconnect [like this example in wagmi](https://github.com/wevm/wagmi/pull/3784/files/8c213576b671b1c7b033f2f9e1188bcbd5068f82#diff-b05598fc1a520e04372175e02a71b5110937ac12bb448dd939b163f5ad8ee463R65)
- on the extension side, `ExtensionCoinbaseWalletProvider extends CoinbaseWalletProvider` so disconnect can be called now with v3 and also when the extension upgrades to SDK v4

locally

### DIFF
--- a/packages/wallet-sdk/src/core/type/ProviderInterface.ts
+++ b/packages/wallet-sdk/src/core/type/ProviderInterface.ts
@@ -22,11 +22,22 @@ interface ProviderConnectInfo {
 
 // properties explicitly required by EIP-1193 spec
 // + extends EventEmitter per spec recomendation
-export interface ProviderInterface extends EventEmitter {
+interface MinimalProviderInterface extends EventEmitter {
   request<T>(args: RequestArguments): Promise<T>;
   on(event: 'connect', listener: (info: ProviderConnectInfo) => void): this;
   on(event: 'disconnect', listener: (error: ProviderRpcError) => void): this;
   on(event: 'chainChanged', listener: (chainId: string) => void): this;
   on(event: 'accountsChanged', listener: (accounts: string[]) => void): this;
   on(event: 'message', listener: (message: ProviderMessage) => void): this;
+}
+
+//
+export interface ProviderInterface extends MinimalProviderInterface {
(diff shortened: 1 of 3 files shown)


packages/wallet-sdk/src/core/type/ProviderInterface.ts
```diff
@@ -20,10 +20,10 @@ interface ProviderConnectInfo {
   readonly chainId: string;
 }
 
-// properties explicitly required by EIP-1193 spec
-// + extends EventEmitter per spec recomendation
 export interface ProviderInterface extends EventEmitter {
   request<T>(args: RequestArguments): Promise<T>;
+  disconnect(): Promise<void>;
+  setAppInfo?(appName: string, appLogoUrl: string | null): void; // legacy extension only
```
bangtoven, line 26: let's remove this
```suggestion
```

bangtoven (commented):
@nateReiners

packages/wallet-sdk/src/core/type/ProviderInterface.ts
nateReiners: Okay but then ts is going to complain about [this](https://github.com/coinbase/coinbase-wallet-sdk/pull/1166/files/f318595ddbc737e98b84cc98a8109eb3dd626045#diff-39f5c3cd1d01da4693d4d2bffc017f605a8df261c485fb878c5ae723b0779a5aL53)

So should I just create a `LegacyProviderInterface` again that extends `ProviderInterface` and adds `setAppInfo`?

bangtoven: nah. just check whether `setAppInfo` function is available within the function. we don't need to expose it in the interface. we just use it internally on one location

bangtoven (approved):

PR #1169, opened 2024-04-11
Revamping 4.0 structure

fan-zhang-sv:


packages/wallet-sdk/src/CoinbaseWalletProvider.ts
```diff
@@ -23,6 +24,7 @@ interface ConstructorOptions {
   appLogoUrl?: string | null;
   appChainIds: number[];
   smartWalletOnly: boolean;
+  injectedSigner?: Signer;
```
bangtoven, line 27: injectedSigner sounds ok.
how about `designatedSigner` or `preselectedSigner`?

packages/wallet-sdk/src/core/type/index.ts
```diff
@@ -1,5 +1,17 @@
 // Copyright (c) 2018-2024 Coinbase, Inc. <https://www.coinbase.com/>
 
+import { ProviderInterface } from './ProviderInterface';
+
+interface LegacyProviderInterface extends ProviderInterface {
+  setAppInfo?(appName: string, appLogoUrl: string | null): void;
+}
```
bangtoven, line 7: let's just remove this. `setAppInfo` doesn't need to be exposed. it's being used internally only at one location. we can check it's availability directly there

packages/wallet-sdk/src/sign/SignRequestHandler.ts
```diff
@@ -114,6 +112,14 @@ export class SignRequestHandler implements RequestHandler {
   }
 
   async onDisconnect() {
+    this._signer = undefined;
     await this.signerConfigurator.onDisconnect();
   }
+
+  private async initSignerIfNeeded(): Promise<Signer> {
```
bangtoven, line 119: not sure this is valid ts, but can we do something like
```suggestion
  private get signer(): Promise<Signer> {
```
returning a promise?

packages/wallet-sdk/src/CoinbaseWalletProvider.ts
```diff
@@ -103,6 +97,10 @@ export class CoinbaseWalletProvider extends EventEmitter implements ProviderInte
     });
   }
 
+  public get isCoinbaseWallet() {
+    return true;
+  }
```
bangtoven, line 102: we could do something like
```suggestion
  readonly isCoinbaseWallet = true
```

packages/wallet-sdk/src/CoinbaseWalletSDK.ts
```diff
@@ -87,33 +65,27 @@ export class CoinbaseWalletSDK {
     return walletLogo(type, width);
   }
 
-  private get walletExtension(): LegacyProviderInterface | undefined {
-    return window.coinbaseWalletExtension;
-  }
+  private grabInjectedProvider(): ProviderInterface | undefined {
```
bangtoven, line 68: how about having this as a separate util function?
doesn't seem like it uses any internal methods of this class anyway.
and, what would be a better name for this?

packages/wallet-sdk/src/sign/SignRequestHandler.ts
```diff
@@ -58,29 +65,20 @@ export class SignRequestHandler implements RequestHandler {
...
-      // when the wallet link QR code url is requested
-      this.signerConfigurator.initSigner();
-
-      const ethAddresses = await this.signerConfigurator.signer?.handshake();
+      const signer = await this.initSignerIfNeeded();
+      const ethAddresses = await signer.handshake();
       if (Array.isArray(ethAddresses)) {
-        if (this.signerConfigurator.signerType === 'walletlink') {
-          this.popupCommunicator.walletLinkQrScanned();
+        if (this._signer instanceof WLSigner) {
```
bangtoven, line 72: why not use local `signer`?

packages/wallet-sdk/src/sign/walletlink/relay/connection/WalletLinkConnection.ts
```diff
@@ -354,7 +354,7 @@ export class WalletLinkConnection {
       JSON.stringify({
         ...unencryptedData,
         origin: location.origin,
-        relaySource: window.coinbaseWalletExtension ? 'injected_sdk' : 'sdk',
```
bangtoven, line 357: can we keep this? we just want to check `coinbaseWalletExtension` exists

packages/wallet-sdk/src/sign/scw/transport/PopUpConfigurator.ts
```diff
@@ -1,101 +0,0 @@
-import { LIB_VERSION } from '../../../version';
```
bangtoven, line 1: 👋

packages/wallet-sdk/src/sign/scw/transport/ConfigMessage.ts
```diff
@@ -3,24 +3,40 @@ import { Message } from ':core/communicator/Message';
...
 
-export enum ClientConfigEventType {
-  SelectConnectionType = 'selectConnectionType',
-  DappOriginMessage = 'dappOriginMessage',
-  WalletLinkUrl = 'walletLinkUrl',
-  WalletLinkQrScanned = 'walletLinkQrScanned',
+export type ConfigEventType =
+  | PopupSetupEventType
+  | SignerConfigEventType
+  | WalletLinkConfigEventType;
```
bangtoven, line 14: nice structure.
let's revisit the names though.
would be nice if they look consistent, and easy to understand who's sender and receiver and logical sequence of them.
take a look at things like "ssl handshake process". might give some inspirations?

and maybe the whole structure of this type 
image
can it be just
type: 'config'
message: ConfigEventType
params?: unknown
?

packages/wallet-sdk/src/sign/scw/SCWSigner.ts
```diff
@@ -111,7 +105,7 @@ export class SCWSigner implements Signer {
     await this.keyManager.clear();
   }
 
-  private tryLocalHandling<T>(request: RequestArguments): T | undefined {
+  private async tryLocalHandling<T>(request: RequestArguments): Promise<T | undefined> {
```
bangtoven, line 108: looking good.

packages/wallet-sdk/src/sign/SignerConfigurator.ts
```diff
@@ -41,129 +48,114 @@ export class SignerConfigurator {
...
+        resolve(signerType);
+      };
+      this.popupCommunicator.rejectSignerSelection = reject;
 
-    this.signer = extensionSigner;
+      this.popupCommunicator.connect({
+        onboarding: true,
+        smartWalletOnly: this.smartWalletOnly,
+        isExtSignerAvailable: Boolean(this.injectedSigner),
+      });
```
bangtoven, line 113: 1. where do we use this with different options? do we set `onboarding: false` somewhere?
2. do we need all of these? `isExtSignerAvailable` can be checked on the popup itself.

packages/wallet-sdk/src/sign/SignerConfigurator.ts
```diff
@@ -26,11 +33,11 @@ export class SignerConfigurator {
   private popupCommunicator: PopUpCommunicator;
   private updateListener: SignRequestHandlerListener;
 
-  private signerTypeStorage = new ScopedLocalStorage('CBWSDK', 'SignerConfigurator');
-  private signerTypeSelectionResolver: ((signerType: string) => void) | undefined;
+  private injectedSigner: Signer | undefined;
+  private initiatedWLSigner: WLSigner | undefined;
```
bangtoven, line 37: seems we need to distinguish
initiate v.s. select a signer

we expect only one signer to be selected at a time.
but at the same time, we might be initiating more than one of them, e.g. initiating WL for qrcode.

packages/wallet-sdk/src/CoinbaseWalletProvider.ts
fan-zhang-sv: i tried to avoid this, cause user may not choose this signer, cause even Extension inject this signer in, user still will go to SCW popup and choose one, and that could be a different signer

packages/wallet-sdk/src/sign/SignerConfigurator.ts
```diff
@@ -26,11 +33,11 @@ export class SignerConfigurator {
...
 
-  signerType: string | null;
-  signer: Signer | undefined;
+  private signerTypeStorage = new ScopedLocalStorage('CBWSDK', 'SignerConfigurator');
+  private resolveSelectSignerType?: (_: SignerType) => void;
```
bangtoven, line 40: can we remove this?
seems it does only two things
image

- store: it could be directly called when `this.resolveSelectSignerType?.()` is being called
- resolve when it's done: doesn't seem like it's observed by anyone?

packages/wallet-sdk/src/sign/SignRequestHandler.ts
fan-zhang-sv: yea.. tried but we cant do this for getter in TS yet

packages/wallet-sdk/src/sign/SignerConfigurator.ts
fan-zhang-sv: i dont think we can tho, this resolver is being called at 2 places
1. SCW&Extension: at handleSignerConfigMessage() when receiving SignerTypeSelected
2. WalletLink: at L159, when WLAdaptor update status

bangtoven: what do they await for? why do we have a separate resolver?

packages/wallet-sdk/src/CoinbaseWalletProvider.ts
bangtoven: you know what.
i think we are about to make the same mistake as the previous SDKs.
if extension needs customization, they should do it via creating a subclass or something, instead of passing some constructor params.

i think we should just remove this option, and maybe introduce it later on the extension side.

packages/wallet-sdk/src/sign/SignerConfigurator.ts
fan-zhang-sv: it's the same resolver for selecting a signerType, but we call it at different place

1. SCW and extension is considered selected, when user click the button from SCW popup
2. WalletLink is considered selected after WalletLink Adaptor established connection

packages/wallet-sdk/src/CoinbaseWalletProvider.ts
fan-zhang-sv: oh wow, you're right

packages/wallet-sdk/src/sign/SignRequestHandler.ts
```diff
@@ -114,6 +112,14 @@ export class SignRequestHandler implements RequestHandler {
...
   async onDisconnect() {
+    this._signer = undefined;
     await this.signerConfigurator.onDisconnect();
   }
+
+  private async initSignerIfNeeded(): Promise<Signer> {
+    if (this._signer) return this._signer;
+
+    this._signer = await this.signerConfigurator.initSigner();
+    return this._signer;
```
bangtoven:
```suggestion
    return this.signerConfigurator.initSigner().then(signer => {this.signer = signer; return signer});
```

packages/wallet-sdk/src/sign/SignRequestHandler.ts
bangtoven: @fan-zhang-sv try something like this https://github.com/coinbase/coinbase-wallet-sdk/pull/1169/files/f051aadda9be52512163b2eeff80c37448116c50#r1561853445
as long as you don't use `await`, i don't think ts has any issues with having computed getter with promise as return type.

packages/wallet-sdk/src/sign/SignerConfigurator.ts
bangtoven: sounds like we just need common interface for selection updates on all those signer types?

packages/wallet-sdk/src/sign/SignRequestHandler.ts
fan-zhang-sv: yea it works,

but when we try to use this signer, we'd either:
1. `(await (await this.signer).request(request);)`
2. or do `const signer = await this.signer` beforehand, which is similar to what we have now

still pretty troublesome? plus the getter become very hard to read
```
  private get signer(): Promise<Signer> {
    if (this._signer) return Promise.resolve(this._signer);

    return this.signerConfigurator.initSignerFromUserSelection().then((signer) => {
      this._signer = signer;
      return signer;
    });
  }
```

packages/wallet-sdk/src/CoinbaseWalletProvider.ts
fan-zhang-sv: removed

packages/wallet-sdk/src/CoinbaseWalletSDK.ts
fan-zhang-sv: udpated

packages/wallet-sdk/src/sign/walletlink/relay/connection/WalletLinkConnection.ts
fan-zhang-sv: added back

packages/wallet-sdk/src/sign/SignRequestHandler.ts
bangtoven: gotcha. not so cleaner after all.

packages/wallet-sdk/src/sign/SignerConfigurator.ts
fan-zhang-sv: good call
1. removed `isExtSignerAvailable`, will let SCW FE to detect extension
2. i tend to keep onboarding: true in the search param tho, so it can explicitly tell FE to redirect to onboarding flow. lmk wdyt

packages/wallet-sdk/src/sign/SignerConfigurator.ts
fan-zhang-sv: updated naming in this area!

packages/wallet-sdk/src/sign/SignerConfigurator.ts
fan-zhang-sv: hmm wdym by common interface
i tried simplified this area a bit:
1. broke down this resolver to only take the resolve
2. created a private function to handle signerTypeSelected

packages/wallet-sdk/src/CoinbaseWalletSDK.ts
```diff
@@ -83,28 +65,32 @@ export class CoinbaseWalletSDK {
...
-  private get coinbaseBrowser(): ProviderInterface | undefined {
-    try {
-      // Coinbase DApp browser does not inject into iframes so grab provider from top frame if it exists
-      const ethereum = window.ethereum ?? window.top?.ethereum;
-      if (!ethereum) {
-        return undefined;
-      }
+  const extension = window.coinbaseWalletExtension;
+  if (extension) {
+    if ('shouldUseSigner' in extension && extension.shouldUseSigner) return undefined;
```
bangtoven, line 82: let’s make fetchInjectedProvider stateless. we don’t want any side effects (calling set***)
i know the code would look bit ugly, but passing appName and logo url seems not ideal.

how about:
```suggestion
  if (extension && !('shouldUseSigner' in extension && extension.shouldUseSigner)) {
    return extension;
```

and we might consider moving this to core/utils or somewhere there, with name something like fetchCoinbaseInjectedProvider to make it clear that it's to grab a provider injected by cb?

packages/wallet-sdk/src/core/communicator/CrossDomainCommunicator.ts
```diff
@@ -13,13 +13,13 @@ export abstract class CrossDomainCommunicator {
     this._connected = value;
   }
 
-  protected abstract onConnect(): Promise<void>;
+  protected abstract onConnect(params?: Record<string, unknown>): Promise<void>;
```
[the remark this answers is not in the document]
bangtoven: please take a look at my last comment.
i don't think we need any changes on CDC file.

packages/wallet-sdk/src/core/type/index.ts
```diff
@@ -1,5 +1,13 @@
 // Copyright (c) 2018-2024 Coinbase, Inc. <https://www.coinbase.com/>
 
+import { ProviderInterface } from './ProviderInterface';
+
+export interface LocalWindow extends Window {
+  top: LocalWindow;
```
bangtoven, line 6: should it be LocalWindow or Window?

packages/wallet-sdk/src/sign/walletlink/relay/connection/WalletLinkConnection.ts
```diff
@@ -354,7 +354,10 @@ export class WalletLinkConnection {
...
+        relaySource:
+          'coinbaseWalletExtension' in window && window.coinbaseWalletExtension
+            ? 'injected_sdk'
+            : 'sdk',
```
bangtoven, line 360: 👍

packages/wallet-sdk/src/sign/scw/transport/PopUpCommunicator.ts
```diff
@@ -56,59 +99,42 @@ export class PopUpCommunicator extends CrossDomainCommunicator {
...
+      case PopupSetupEventType.PopupUnload:
+        this.disconnect();
+        break;
+    }
   }
 
   // Window Management
 
-  private openFixedSizePopUpWindow() {
+  private openFixedSizePopUpWindow(params?: Record<string, unknown>) {
```
bangtoven, line 129: look at my previous comments on the params for this

packages/wallet-sdk/src/sign/scw/transport/ConfigMessage.ts
bangtoven: let's revisit this. how can we make this more organized? 
don't have a solution myself for now, but it doesn't look optimal anyway

bangtoven (commented):
@fan-zhang-sv

packages/wallet-sdk/src/core/type/index.ts
fan-zhang-sv: updated

bangtoven (dismissed):

cb-heimdall:
Review Error for bangtoven @ 2024-04-16 22:18:10 UTC

packages/wallet-sdk/src/core/type/index.ts
```diff
@@ -1,5 +1,33 @@
...
 
+import { ProviderInterface } from './ProviderInterface';
+
+export interface Window {
+  top: Window;
+  ethereum?: ProviderInterface;
+  coinbaseWalletExtension?: ProviderInterface;
+}
+
+export function fetchCoinbaseInjectedProvider(
```
bangtoven, line 11: let's move this to https://github.com/coinbase/coinbase-wallet-sdk/blob/master/packages/wallet-sdk/src/core/eip1193Utils.ts (we might rename this file though)

bangtoven (commented):
@fan-zhang-sv

bangtoven (approved):

PR #1171, opened 2024-04-16
SignRequestHandler to take optional `keysUrl`, clean up constructor options

bangtoven:


packages/wallet-sdk/src/sign/SignRequestHandler.ts
```diff
@@ -4,16 +4,13 @@ import { SignRequestHandlerListener } from './UpdateListenerInterface';
...
 
-interface SignRequestHandlerOptions {
-  appName: string;
-  appLogoUrl?: string | null;
-  appChainIds: number[];
-  smartWalletOnly: boolean;
+type SignRequestHandlerOptions = ConstructorOptions & {
   updateListener: SignRequestHandlerListener;
-}
+  keysUrl?: string;
```
fan-zhang-sv, line 12: wonder what would we use this for? 
also nit: i dont have better names in mind, but think keysUrl might not be really intuitive

PR #1176, opened 2024-04-16
Deprecate subscription

bangtoven:
### _Summary_

Dapps can still use our provider to make RPC calls for subscription using external tools like the ones we used to import.


nateReiners (approved):
amazing

PR #1195, opened 2024-04-22
Fix v4 example with wagmi

danpopenko:
### _Summary_

Its looks like now chainIds not exist, but chainId exist

`Object literal may only specify known properties, but 'chainIds' does not exist in type '{ appName: string; appLogoUrl?: string | null; darkMode?: boolean; linkAPIUrl?: string; uiConstructor?: (options: RelayUIOptions) => RelayUI; diagnosticLogger?: any; overrideIsMetaMask?: boolean; overrideIsCoinbaseWallet?: boolean; ... 5 more ...; reloadOnDisconnect?: boolean | undefined; }'. Did you mean to write 'chainId'?`

### DIFF
--- a/packages/wallet-sdk/docs/v4_with_wagmi.md
+++ b/packages/wallet-sdk/docs/v4_with_wagmi.md
@@ -36,7 +36,7 @@ This quickstart guide walks through building a dapp for Coinbase smart wallet us
 
    export const config = createConfig({
      chains: [baseSepolia],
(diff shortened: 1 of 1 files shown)


nateReiners:
Hi @danpopenko it looks like you might have been following the v4 beta walkthrough but using v3 which explains the options discrepancy. 

In v3 we have a `chainId` option, in v4 beta.6 we have a `chainIds` option, but that interface is changing again soon in v4 beta.8. I will update the docs accordingly after beta.8 is released.

PR #1206, opened 2024-04-24
cleanup SignRequestHandler

bangtoven:

### DIFF
--- a/packages/wallet-sdk/src/CoinbaseWalletProvider.ts
+++ b/packages/wallet-sdk/src/CoinbaseWalletProvider.ts
@@ -2,12 +2,7 @@
 import EventEmitter from 'eventemitter3';
 
 import { standardErrors } from './core/error';
-import {
-  ConstructorOptions,
-  ProviderInterface,
-  ProviderRpcError,
-  RequestArguments,
-} from './core/provider/interface';
+import { ConstructorOptions, ProviderInterface, RequestArguments } from './core/provider/interface';
 import { checkErrorForInvalidRequestArgs, fetchRPCRequest } from './core/provider/util';
 import { AddressString, Chain } from './core/type';
 import { areAddressArraysEqual, prepend0x, showDeprecationWarning } from './core/util';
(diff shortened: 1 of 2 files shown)


packages/wallet-sdk/src/CoinbaseWalletProvider.ts
```diff
@@ -57,8 +42,16 @@ export class CoinbaseWalletProvider extends EventEmitter implements ProviderInte
...
+    sign: (request: RequestArguments) => {
+      if (request.method === 'eth_requestAccounts') {
+        if (this.connected) return this.accounts;
+      } else if (!this.connected) {
+        throw standardErrors.provider.unauthorized(
+          "Must call 'eth_requestAccounts' before other methods"
+        );
+      }
+      return this.signRequestHandler.handleRequest(request);
+    },
```
fan-zhang-sv, line 54: it kinda took me quite some time to understand this block tho, i'm leaning towards to not have `else if`
```suggestion
    sign: (request: RequestArguments) => {
      if (request.method === 'eth_requestAccounts') {
        if (this.connected) return this.accounts;
      };
      if (!this.connected && request.method !== 'eth_requestAccounts') {
        throw standardErrors.provider.unauthorized(
          "Must call 'eth_requestAccounts' before other methods"
        );
      };
      return this.signRequestHandler.handleRequest(request);
    },
```

packages/wallet-sdk/src/sign/SignRequestHandler.ts
```diff
@@ -20,34 +16,14 @@ export class SignRequestHandler {
...
-      const signer = await this.useSigner();
-      if (accounts.length <= 0) {
-        throw standardErrors.provider.unauthorized(
-          "Must call 'eth_requestAccounts' before other methods"
-        );
-      }
-
-      return await signer.request(request);
-    } catch (err) {
-      if ((err as { code?: unknown })?.code === standardErrorCodes.provider.unauthorized) {
```
fan-zhang-sv, line 39: this unauthorized is for handling a few edge cases, including where user disconnected from dapp side, but SCW is still at "connected" state. think we cant remove it?

packages/wallet-sdk/src/CoinbaseWalletProvider.ts
bangtoven: linter can't think like human

packages/wallet-sdk/src/sign/SignRequestHandler.ts
```diff
@@ -20,21 +17,15 @@ export class SignRequestHandler {
...
-          "Must call 'eth_requestAccounts' before other methods"
-        );
+      if (request.method === 'eth_requestAccounts') {
+        const accounts = await signer.handshake();
+        this.updateListener.onConnect();
+        return accounts;
       }
-
-      return await signer.request(request);
+      return signer.request(request);
```
fan-zhang-sv, line 28: we might need to await here, otherwise the error won't get caught here, because we returned promise object and ended here already
```suggestion
      return await signer.request(request);
```

fan-zhang-sv (approved):

PR #1208, opened 2024-04-25
Clean up SignRequestHandler

bangtoven:
### _Summary_

- separate out eth_requestAccounts
- deprecate SignRequestHandler
- repurpose SignerConfigurator as SignHandler

unit tests


packages/wallet-sdk/src/CoinbaseWalletProvider.ts
```diff
@@ -42,17 +42,27 @@ export class CoinbaseWalletProvider extends EventEmitter implements ProviderInte
...
-          throw standardErrors.provider.unauthorized(
-            "Must call 'eth_requestAccounts' before other methods"
-          );
-        }
+    // eth_requestAccounts
+    handshake: async (_: RequestArguments): Promise<AddressString[]> => {
+      if (this.connected) return this.accounts;
+      const accounts = await this.signHandler.handshake();
+      this.emit('connect', { chainId: prepend0x(this.chain.id.toString(16)) });
+      return accounts;
```
bangtoven, line 50: moved from SignRequestHandler
image

packages/wallet-sdk/src/CoinbaseWalletProvider.ts
```diff
@@ -40,21 +39,31 @@ export class CoinbaseWalletProvider extends EventEmitter implements ProviderInte
...
+        throw standardErrors.provider.unauthorized(
+          "Must call 'eth_requestAccounts' before other methods"
+        );
+      }
+      try {
+        return await this.signHandler.request(request);
+      } catch (error) {
+        const e = error as { code?: number };
+        if (e.code === standardErrorCodes.provider.unauthorized) this.disconnect();
+        throw error;
```
bangtoven, line 61: moved from SignRequestHandler
image

packages/wallet-sdk/src/CoinbaseWalletProvider.ts
fan-zhang-sv: we need to put handshake() into a try block as well

on FE side, user can click sign out from anywhere, including eth_requestAccounts step, when that happens, FE will respond an unauthorized error back to SDK, and we need to catch it here and call disconnect to clean up

packages/wallet-sdk/src/sign/SignerConfigurator.ts
```diff
@@ -1,107 +1,113 @@
...
-  tryRestoringSignerFromPersistedType(): Signer | undefined {
-    const persistedSignerType = this.signerTypeStorage.getItem(SIGNER_TYPE_KEY) as SignerType;
-    if (persistedSignerType) {
-      return this.initSignerFromType(persistedSignerType);
+  async handshake() {
+    if (!this.signer) {
+      this.signer = await this.selectSigner();
     }
-    return undefined;
+    return await this.signer.handshake();
```
fan-zhang-sv, line 51: nit: we dont need to await here, when we dont need to catch error

packages/wallet-sdk/src/sign/SignerConfigurator.ts
```diff
@@ -1,107 +1,113 @@
...
-    const signerType = await this.requestSignerSelection();
-    if (signerType === 'walletlink' && this.walletlinkSigner) {
-      return this.walletlinkSigner;
+  async request(request: RequestArguments) {
+    if (!this.signer) {
+      throw new Error('Signer is not initialized');
     }
-    const signer = this.initSignerFromType(signerType);
-    return signer;
+    return await this.signer.request(request);
```
fan-zhang-sv, line 58: nit: we dont need to await here, when we dont need to catch error

PR #1218, opened 2024-04-25
small fix

fan-zhang-sv:

### DIFF
--- a/packages/wallet-sdk/src/CoinbaseWalletProvider.ts
+++ b/packages/wallet-sdk/src/CoinbaseWalletProvider.ts
@@ -44,6 +44,7 @@ export class CoinbaseWalletProvider extends EventEmitter implements ProviderInte
     // eth_requestAccounts
     handshake: async (_: RequestArguments): Promise<AddressString[]> => {
       if (this.connected) {
+        this.emit('connect', { chainId: prepend0x(this.chain.id.toString(16)) });
         return this.accounts;
       }
       try {
--- a/packages/wallet-sdk/src/sign/SignHandler.ts
+++ b/packages/wallet-sdk/src/sign/SignHandler.ts
@@ -33,7 +33,6 @@ export class SignHandler {
   private readonly storage = new ScopedLocalStorage('CBWSDK', 'SignerConfigurator');
 
   constructor(params: Readonly<SignerConfiguratorOptions>) {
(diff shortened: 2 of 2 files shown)


packages/wallet-sdk/src/CoinbaseWalletProvider.ts
```diff
@@ -44,6 +44,7 @@ export class CoinbaseWalletProvider extends EventEmitter implements ProviderInte
     // eth_requestAccounts
     handshake: async (_: RequestArguments): Promise<AddressString[]> => {
       if (this.connected) {
+        this.emit('connect', { chainId: prepend0x(this.chain.id.toString(16)) });
```
fan-zhang-sv, line 47: turns out we still need this, when dapp reloads and they dont rememebr logged in state, we'd need to alert them

bangtoven, line 47: i thought we didn't emit `connect` for already connected case?

packages/wallet-sdk/src/sign/SignHandler.ts
```diff
@@ -42,6 +41,7 @@ export class SignHandler {
       url: keysUrl ?? CB_KEYS_URL,
       onConfigUpdateMessage: this.handleConfigUpdateMessage.bind(this),
     });
+    this.signer = this.loadSigner();
```
fan-zhang-sv, line 44: init signer needs the metadata, etc assigned

bangtoven, line 44: is there any issue with sequence of these lines?

bangtoven (commented):
@fan-zhang-sv

bangtoven (approved):

packages/wallet-sdk/src/CoinbaseWalletProvider.ts
fan-zhang-sv: the playground testdapp is one example

it loses connected state on reload, and it needs this connect message to enable the rest of functionalities

packages/wallet-sdk/src/sign/SignHandler.ts
fan-zhang-sv: yea, loadSigner reads this.metadata, etc, which was undefined

PR #1220, opened 2024-04-25
updateListener protected

fan-zhang-sv:


packages/wallet-sdk/src/CoinbaseWalletProvider.ts
```diff
@@ -126,7 +126,7 @@ export class CoinbaseWalletProvider extends EventEmitter implements ProviderInte
 
   readonly isCoinbaseWallet = true;
 
-  private readonly updateListener = {
+  protected readonly updateListener = {
```
fan-zhang-sv, line 129: subclass needs to pass this in when replacing SignHandler

bangtoven (approved):

PR #1233, opened 2024-04-26
default test app to dark mode and system prefs

gareys-cb:
### _Summary_

I made dark mode the default

locally

Screenshot 2024-04-26 at 1 40 43 PM


fan-zhang-sv (approved):
lololol ty, but i'll make it switchable later

PR #1244, opened 2024-04-29
#fix appName can be empty string

fan-zhang-sv:
### _Summary_

appName and appLogoUrl can be empty string. Updated handling to make that also defaulting to 'Dapp' and getFavicon()

### DIFF
--- a/packages/wallet-sdk/src/CoinbaseWalletSDK.ts
+++ b/packages/wallet-sdk/src/CoinbaseWalletSDK.ts
@@ -12,31 +12,20 @@ import { getCoinbaseInjectedProvider } from ':util/provider';
 type CoinbaseWalletSDKOptions = Partial<AppMetadata>;
 
 export class CoinbaseWalletSDK {
-  private metadata: CoinbaseWalletSDKOptions;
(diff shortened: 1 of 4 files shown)


packages/wallet-sdk/src/CoinbaseWalletSDK.ts
```diff
@@ -20,7 +20,9 @@ export class CoinbaseWalletSDK {
   }
 
   public makeWeb3Provider(preference: Preference = { options: 'all' }): ProviderInterface {
-    const { appName = 'Dapp', appLogoUrl = getFavicon(), appChainIds = [] } = this.metadata;
+    const { appName = '', appLogoUrl = '', appChainIds = [] } = this.metadata;
+    const safeAppName = appName || 'Dapp';
+    const safeAppLogoUrl = appLogoUrl || getFavicon();
```
bangtoven, line 25: is there cleaner way to do this, without creating safe*** variables?

bangtoven (commented):
@fan-zhang-sv

bangtoven (approved):

PR #1266, opened 2024-05-03
Prototype Communicator test

eric-ships:
### _Summary_

This PR prototypes a simple solution of testing Communicator.

Given the async nature of message passing, this is a prototype of a simple solution to queue up message events and then dispatch them every 200ms.

If this proposal is accepted, the plan would be to keep iterating forward and, ultimately, move these test helpers into its own module.

Unit tests


bangtoven (approved):
❤️

PR #1276, opened 2024-05-06
Improve popup lifecycle management

bangtoven:
### _Summary_

- introduce a dedicated method `postRPCRequest` which manages the pending requests and determines when to call `disconnect` to close the popup
- `postMessage`: no conditional behavior 
- replace binding with const func blocks

- new unit tests
- playground


packages/wallet-sdk/src/core/communicator/Communicator.ts
```diff
@@ -14,59 +14,85 @@ import { standardErrors } from ':core/error';
...
+        await this.post(request);
+        const response: RPCResponseMessage = await this.onMessage(
+          ({ requestId }) => requestId === request.id,
+          { throwWhenPopupClose: true }
+        );
+        return response;
+      })
+      .finally(() => {
+        // unless FE popup supports batch requests, close the popup after each request
+        // this.close();
```
bangtoven, line 51: let's explore

packages/wallet-sdk/src/core/communicator/Communicator.ts
```diff
@@ -14,31 +14,24 @@ import { standardErrors } from ':core/error';
...
 
   constructor(url: string = CB_KEYS_URL) {
     this.url = new URL(url);
-    this.postMessage = this.postMessage.bind(this);
-    this.onMessage = this.onMessage.bind(this);
-    this.disconnect = this.disconnect.bind(this);
   }
 
   /**
-   * Posts a message to the popup window and optionally waits for a response.
```
bangtoven, line 27: `optionally waits for a response` will be handled on the SCWSigner, where we actually make RPC requests and wait for response.

packages/wallet-sdk/src/core/communicator/Communicator.ts
```diff
@@ -14,31 +14,25 @@ import { standardErrors } from ':core/error';
...
```
bangtoven, line 27: no conditional behavior

packages/wallet-sdk/src/CoinbaseWalletProvider.ts
```diff
@@ -176,7 +176,7 @@ export class CoinbaseWalletProvider extends EventEmitter implements ProviderInte
         return new SCWSigner({
           metadata: this.metadata,
           updateListener: this.updateListener,
-          postMessageToPopup: this.communicator.postMessage.bind(this),
```
fan-zhang-sv, line 179: i just realized we were binding this wrong before lol

fan-zhang-sv (approved):

packages/wallet-sdk/src/core/communicator/Communicator.ts
```diff
@@ -96,5 +91,26 @@ export class Communicator {
...
+        .catch(reject)
+        .finally(() => {
+          // then decrement the pending count and disconnect if no more requests
+          this.pendingRPCRequestsCount--;
+          if (this.pendingRPCRequestsCount === 0) {
+            this.disconnect();
+          }
+        });
+    });
+  };
```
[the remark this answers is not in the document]
bangtoven: @fan-zhang-sv 
i don't think this works, as finally block would get called before the potential subsequent call can increase the count.

packages/wallet-sdk/src/core/communicator/Communicator.ts
```diff
@@ -96,5 +110,35 @@ export class Communicator {
...
         return this.popup;
       });
-  }
+  };
+
+  /**
+   * Posts a RPC request to the popup window and waits for a response
+   */
+  postRPCRequest = (request: RPCRequestMessage): Promise<RPCResponseMessage> => {
+    this.pendingRequests.set(request.id, 'created');
```
bangtoven, line 119: dapp "creates" a request that hasn't been awaited yet.
e.g. `const promise = communicator.postRequest(request)`

packages/wallet-sdk/src/core/communicator/Communicator.ts
```diff
@@ -96,5 +110,35 @@ export class Communicator {
...
+
+    return new Promise((resolve, reject) => {
+      // reject if the request was cancelled between creation and awaiting due to disconnect
+      if (!this.pendingRequests.has(request.id)) {
+        reject(standardErrors.provider.userRejectedRequest('Request cancelled before sending'));
+        return;
+      }
```
bangtoven, line 126: non-`awaiting` requests can be cancelled without getting posted to popup, if popup explicitly has sent `PopupUnload` event before it's awaited.

packages/wallet-sdk/src/core/communicator/Communicator.ts
```diff
@@ -96,5 +110,35 @@ export class Communicator {
...
+
+      // otherwise, post the message and wait for a response
+      this.pendingRequests.set(request.id, 'awaiting');
+      this.postMessage(request)
+        .then(() => this.onMessage<RPCResponseMessage>(({ requestId }) => requestId === request.id))
```
bangtoven, line 131: `onMessage` can be cancelled as well, if popup gets closed while waiting for response. 
image

packages/wallet-sdk/src/core/communicator/Communicator.ts
```diff
@@ -96,5 +110,35 @@ export class Communicator {
...
+        // resolve or reject the outer promise to take any follow-up requests before disconnect
+        .then(resolve)
+        .catch(reject)
+        // then clean up the current request and disconnect if there are no more pending requests
+        .finally(() => {
+          this.pendingRequests.delete(request.id);
+          if (this.pendingRequests.size === 0) {
+            this.disconnect();
+          }
```
bangtoven, line 140: regardless of the result, all `awaiting` requests will eventually hit this block. if there are no pending requests, it can close the popup.

packages/wallet-sdk/src/core/communicator/Communicator.ts
```diff
@@ -96,5 +110,40 @@ export class Communicator {
...
+    this.pendingRequests.set(request.id, 'created');
+
+    return new Promise((resolve, reject) => {
+      // defer the execution the promise until the current call stack clears
+      Promise.resolve().then(() => {
+        // reject if the request was cancelled between creation and awaiting due to PopupUnload
+        if (!this.pendingRequests.has(request.id)) {
+          reject(standardErrors.provider.userRejectedRequest('Request cancelled before sending'));
+          return;
+        }
```
bangtoven, line 128: non-`awaiting` requests can be cancelled without getting posted to popup, if popup explicitly has sent `PopupUnload` event before it's awaited.

packages/wallet-sdk/src/core/communicator/Communicator.ts
```diff
@@ -96,5 +110,40 @@ export class Communicator {
...
+
+        // otherwise, post the message and wait for a response
+        this.pendingRequests.set(request.id, 'awaiting');
+        this.postMessage(request)
+          .then(() =>
+            this.onMessage<RPCResponseMessage>(({ requestId }) => requestId === request.id)
```
bangtoven, line 134: `onMessage` can be cancelled as well, if popup gets closed while waiting for response. 
image

packages/wallet-sdk/src/core/communicator/Communicator.ts
```diff
@@ -96,5 +110,40 @@ export class Communicator {
...
+          )
+          // resolve or reject the promise to take any follow-up requests if necessary before disconnecting
+          .then(resolve)
+          .catch(reject)
+          // then clean up the current request and disconnect if there are no more pending requests
+          .finally(() => {
+            this.pendingRequests.delete(request.id);
+            if (this.pendingRequests.size === 0) {
+              this.disconnect();
+            }
```
bangtoven, line 144: regardless of the result, all `awaiting` requests will eventually hit this block. if there are no pending requests, it can close the popup.

packages/wallet-sdk/src/core/communicator/Communicator.test.ts
```diff
@@ -97,4 +98,122 @@ describe('Communicator', () => {
...
+        postMessageSpy = jest.spyOn(communicator, 'postMessage').mockResolvedValue();
+
+        const promise = communicator.postRPCRequest(request1);
+        expect(postMessageSpy).not.toHaveBeenCalledWith(request1);
+        expect(communicator.disconnect).not.toHaveBeenCalled();
+
+        // close the popup after a delay
+        setTimeout(() => communicator.disconnect(), 100);
+
+        await expect(promise).rejects.toThrow('Request cancelled before response');
```
bangtoven, line 170: this can happen if popup closes before the request promise gets awaited.

packages/wallet-sdk/src/core/communicator/Communicator.test.ts
```diff
@@ -97,4 +98,122 @@ describe('Communicator', () => {
...
+
+      it('should reject if popup disconnects before posting request', async () => {
+        const promise = communicator.postRPCRequest(request1);
+        expect(postMessageSpy).not.toHaveBeenCalledWith(request1);
+        expect(communicator.disconnect).not.toHaveBeenCalled();
+
+        // close the popup
+        communicator.disconnect();
+
+        await expect(promise).rejects.toThrow('Request cancelled before sending');
```
bangtoven, line 156: this can happen if popup closes before the request promise gets awaited.

packages/wallet-sdk/src/core/communicator/Communicator.test.ts
bangtoven, line 170: this happens if popup closes after request promise starts execution, but before receiving response.
e.g. user manually closes the popup

packages/wallet-sdk/src/core/communicator/Communicator.ts
```diff
@@ -96,5 +110,40 @@ export class Communicator {
         if (!this.popup) throw standardErrors.rpc.internal();
         return this.popup;
       });
-  }
+  };
+
+  /**
```
bangtoven, line 115: there are some lines that are quite subtle.
wish there's cleaner solution for this 😞

packages/wallet-sdk/src/core/communicator/Communicator.ts
```diff
@@ -96,5 +110,40 @@ export class Communicator {
...
+   * Posts a RPC request to the popup window and waits for a response
+   */
+  postRPCRequest = (request: RPCRequestMessage): Promise<RPCResponseMessage> => {
+    this.pendingRequests.set(request.id, 'created');
+
+    return new Promise((resolve, reject) => {
+      // defer the execution until the current call stack clears
```
bangtoven, line 122: to cover back to back reqeusts
```
const response1 = await communicator.postRPCRequest(request1);
const response2 = await communicator.postRPCRequest(request2);
```

packages/wallet-sdk/src/core/communicator/Communicator.ts
```diff
@@ -96,5 +110,40 @@ export class Communicator {
...
+            this.onMessage<RPCResponseMessage>(({ requestId }) => requestId === request.id)
+          )
+          // resolve or reject the promise to take any follow-up requests if necessary before disconnecting
+          .then(resolve)
+          .catch(reject)
+          // then clean up the current request and disconnect if there are no more pending requests
+          .finally(() => {
+            this.pendingRequests.delete(request.id);
+            if (this.pendingRequests.size === 0) {
+              this.disconnect();
```
eric-ships, line 143: Is it always desired behavior to have this?

**Scenario: Onboarding**
When the user is onboarding and response to the `eth_requestAccounts`, the pop-up should not close once the response is done. 

The pop-up should persist and display the onboarding success screen to link to Pano or fund.

packages/wallet-sdk/src/sign/util.ts
```diff
@@ -28,7 +28,8 @@ export async function fetchSignerType(params: {
     event: 'selectSignerType',
     data: params.preference,
   };
-  const { data } = await communicator.postMessage(request);
+  await communicator.postMessage(request);
+  const { data } = await communicator.onMessage(({ requestId }) => requestId === request.id);
```
eric-ships, line 32: Is there a possible, but very unlikely, race condition between these two where the response may come back faster than the communicator starts to await?

A way around this could be to start listening first.

This is existing behavior, and don't think it's a big deal either way.

packages/wallet-sdk/src/core/communicator/Communicator.test.ts
```diff
@@ -97,4 +98,122 @@ describe('Communicator', () => {
...
+    const simulateAsyncCall = () => new Promise((resolve) => setTimeout(resolve, 10));
+
+    let postMessageSpy: jest.SpyInstance;
+    let disconnectSpy: jest.SpyInstance;
+
+    beforeEach(() => {
+      queueMessageEvent(popupLoadedMessage);
+
+      disconnectSpy = jest.spyOn(communicator, 'disconnect');
+      postMessageSpy = jest
```
eric-ships, line 114: If this test is for Communicator, then Communicator would be the system under test and the methods should avoid being mocked.

High-level options:
- Do nothing: having these tests are better than none
- Mock underlying functions and verify against those calls
- Abstract out `disconnect` and `postMessage` logic and call this `Communicator` something like `MessageHandler` (or keep it the same and abstract out logic)

eric-ships (commented):
Generally aligned with approach - the key question is about the SDK closing the pop-up when the FE may still want to persist it

packages/wallet-sdk/src/core/communicator/Communicator.ts
eric-ships: I'd like to better understand this comment

packages/wallet-sdk/src/core/communicator/Communicator.ts
```diff
@@ -14,31 +27,27 @@ import { standardErrors } from ':core/error';
...
-    this.disconnect = this.disconnect.bind(this);
   }
 
   /**
-   * Posts a message to the popup window and optionally waits for a response.
+   * Posts a message to the popup window
    */
-  async postMessage<M extends Message>(request: Message): Promise<M> {
+  postMessage = async (message: Message) => {
     const popup = await this.waitForPopupLoaded();
```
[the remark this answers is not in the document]
bangtoven: i believe the intention behind `waitForPopupLoaded ` was to 
- get the pointer to popup that's returned from `this.open...`
- and wait for the popup to send `PopupLoaded` and notify sdk that it's ready to proceed

packages/wallet-sdk/src/core/communicator/Communicator.ts
```diff
@@ -96,5 +110,40 @@ export class Communicator {
...
+        this.postMessage(request)
+          .then(() =>
+            this.onMessage<RPCResponseMessage>(({ requestId }) => requestId === request.id)
+          )
+          // resolve or reject the promise to take any follow-up requests if necessary before disconnecting
+          .then(resolve)
+          .catch(reject)
+          // then clean up the current request and disconnect if there are no more pending requests
+          .finally(() => {
+            this.pendingRequests.delete(request.id);
```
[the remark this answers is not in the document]
bangtoven: i don't think we have those kinds of requests that we send to wallet at the moment.
eth_requestAccounts and all signing requests expect one response.

packages/wallet-sdk/src/core/communicator/Communicator.ts
bangtoven: good point. thanks for sharing the scenarios. that helps me understand the issue better.

this line should be revisited. 
instead of calling `this.disconnect()` here (which means both sdk and popup can initiate this), `this.pendingRequests.size === 0` check should happen inside the `disconnect` method which should be triggered only by `PopupUnload` event from popup.
will take a closer look today.

packages/wallet-sdk/src/sign/util.ts
bangtoven: will address.

packages/wallet-sdk/src/core/communicator/Communicator.test.ts
bangtoven: agreed 100%. 
for now, i will just leave it like this as `postMessage` method is fairly simple.
image
will revisit once we have better async behavior on this pr first.

bangtoven:
turned out that `closePopup` call on the sdk side had no effect, as popup had has been responsible for closing itself whole time.
sdk shouldn't interfere in.
popup is taking care of it all.
closing this pr, replacing it with https://github.com/coinbase/coinbase-wallet-sdk/pull/1279

PR #1282, opened 2024-05-09
protected initSigner

fan-zhang-sv:


bangtoven:
is this for wallet to inherit the provider?
if so, let's revisit.
1. i don't think this is necessary for wallet to inherit
2. even if it's needed, i'd rather create a separate provider class on wallet side, instead of trying to generalize the behavior of sdk's provider. 
cc @fan-zhang-sv

fan-zhang-sv:
sg will revisit this when developing for wallet

PR #1260, opened 2024-05-02
Clean up PopupCommunicator

bangtoven:
### _Summary_

- consolidate CrossDomainCommunicator and PopUpCommunicator
- introduce `onMessage` to listen for individual event, deprecating global event handling using `requestMap`
- let `postMessage` take care of popup's availability, instead of requiring `.connect()` to be called beforehand
- `SignHandler` to extend Communicator instead of having instance of it

### _How did you test your changes?_

- unit tests
- playground

PR #1294, opened 2024-05-19
Fix popup blocked on Safari

arjun-dureja:
### _Summary_

- Safari was blocking the popup from opening whenever a request was sent. This is because the request message is being constructed before the popup is opened in `SCWSigner`. Safari blocks it because it is not opened immediately after a button is clicked.
  - This PR updates the `request` method so that it first opens the popup and then constructs the request. This is how it was done before [this PR](https://github.com/coinbase/coinbase-wallet-sdk/pull/1260)

- Manually:

| Before | After |
| ------------- | ------------- |
| <video src="https://github.com/coinbase/coinbase-wallet-sdk/assets/16995513/619d30b6-ca27-4401-b979-a462a2935a0a">  | <video src="https://github.com/coinbase/coinbase-wallet-sdk/assets/16995513/bf69b157-c3d6-41d9-81fa-cee43cc02b35">|

### DIFF
--- a/packages/wallet-sdk/src/core/communicator/Communicator.ts
+++ b/packages/wallet-sdk/src/core/communicator/Communicator.ts
@@ -87,17 +87,17 @@ export class Communicator {
       });
   };
 
-    /**
+  /**
    * Closes the popup, rejects all requests and clears the listeners
    */
-    private disconnect = () => {
-      closePopup(this.popup);
-      this.popup = null;
-  
-      this.listeners.forEach(({ reject }, listener) => {
-        reject(standardErrors.provider.userRejectedRequest('Request rejected'));
-        window.removeEventListener('message', listener);
-      });
-      this.listeners.clear();
-    };
+  private disconnect = () => {
+    closePopup(this.popup);
+    this.popup = null;
+
+    this.listeners.forEach(({ reject }, listener) => {
+      reject(standardErrors.provider.userRejectedRequest('Request rejected'));
+      window.removeEventListener('message', listener);
+    });
+    this.listeners.clear();
+  };
 }
--- a/packages/wallet-sdk/src/core/communicator/Communicator.ts
+++ b/packages/wallet-sdk/src/core/communicator/Communicator.ts
@@ -63,6 +63,20 @@ export class Communicator {
   };
 
   /**
+   * Closes the popup, rejects all requests and clears the listeners
+   */
+  private disconnect = () => {
+    closePopup(this.popup);
+    this.popup = null;
+
+    this.listeners.forEach(({ reject }, listener) => {
+      reject(standardErrors.provider.userRejectedRequest('Request rejected'));
+      window.removeEventListener('message', listener);
+    });
+    this.listeners.clear();
+  };
+
+  /**
    * Waits for the popup window to fully load and then sends a version message.
    */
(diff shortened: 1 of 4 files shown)


packages/wallet-sdk/src/core/communicator/Communicator.ts
```diff
@@ -62,24 +62,10 @@ export class Communicator {
     });
   };
 
-  /**
-   * Closes the popup, rejects all requests and clears the listeners
-   */
-  private disconnect = () => {
```
wilsoncusack, line 68: why remove this?

packages/wallet-sdk/src/core/communicator/Communicator.ts
```diff
@@ -62,24 +62,10 @@ export class Communicator {
     });
   };
 
-  /**
```
bangtoven, line 65: we don't need to move this

packages/wallet-sdk/src/core/communicator/Communicator.ts
```diff
@@ -62,24 +62,10 @@ export class Communicator {
...
-      window.removeEventListener('message', listener);
-    });
-    this.listeners.clear();
-  };
-
   /**
    * Waits for the popup window to fully load and then sends a version message.
    */
-  private waitForPopupLoaded = async (): Promise<Window> => {
+  waitForPopupLoaded = async (): Promise<Window> => {
```
bangtoven, line 68: should keep this as private

packages/wallet-sdk/src/sign/scw/SCWSigner.ts
```diff
@@ -22,16 +22,19 @@ type SwitchEthereumChainParam = [
...
 export class SCWSigner implements Signer {
   private readonly metadata: AppMetadata;
+  private readonly waitForPopupLoaded: () => Promise<Window>;
   private readonly postMessageToPopup: (_: RPCRequestMessage) => Promise<RPCResponseMessage>;
   private readonly keyManager: SCWKeyManager;
   private readonly stateManager: SCWStateManager;
 
   constructor(params: {
     metadata: AppMetadata;
+    waitForPopupLoaded: () => Promise<Window>;
```
bangtoven, line 32: if it needs to access more than one methods of Communicator instance, then let's just pass comm.
or, provider might just open it for scwsigner instead.

bangtoven (commented):
thanks for opening PR up. @arjun-dureja 
happy to pair if you are down.

wilsoncusack:
@bangtoven for sake of speed, maybe you should take over PR with the changes you want? Not sure what else @arjun-dureja has on plate today

packages/wallet-sdk/src/sign/util.ts
```diff
@@ -48,6 +48,7 @@ export function createSigner(params: {
       return new SCWSigner({
         metadata,
         updateListener,
+        waitForPopupLoaded: communicator.waitForPopupLoaded,
```
bangtoven, line 51: let's just pass communicator instance

packages/wallet-sdk/src/core/communicator/Communicator.ts
arjun-dureja: I moved it down since it's the only private method in the class now

bangtoven (approved):

PR #1295, opened 2024-05-20
remove beta & rc tags

nateReiners:
### _Summary_

Remove these beta and rc tags
Screenshot 2024-05-20 at 9 59 08 AM

automerge=true


bangtoven:
hate to say this, but seems we'd need another beta @nateReiners

nateReiners:
> hate to say this, but seems we'd need another beta @nateReiners

ok edited to only remove the rc tag

bangtoven (approved):

PR #1297, opened 2024-05-20
Fix playground example send: don't send to wallets own address 

nateReiners:
dapps are blocked from calling directly to the smart wallet's address

### DIFF
--- a/apps/testapp/src/components/RpcMethods/shortcut/sendShortcuts.ts
+++ b/apps/testapp/src/components/RpcMethods/shortcut/sendShortcuts.ts
@@ -6,7 +6,7 @@ const ethSendTransactionShortcuts: ShortcutType[] = [
     key: 'Example Tx',
     data: {
       from: ADDR_TO_FILL,
-      to: ADDR_TO_FILL,
+      to: '0xd7876e5dbd2da268a35416ba75d0ce154ed949e3',
       value: '0x0',
     },
   },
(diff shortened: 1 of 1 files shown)


bangtoven (changes requested):
i'd just remove send shortcut

spencerstock (approved):
what address is this?

cb-heimdall:
Review Error for bangtoven @ 2024-05-20 23:55:34 UTC

spencerstock:
> i'd just remove send shortcut

If we can find a suitable hardcoded address I think it's valuable... I've clicked the send shortcut at least 300x to test local changes.

nateReiners:
> what address is this?

Just one of my empty smart wallet test accounts

bangtoven:
@spencerstock why not running a new dapp dedicated to do so and automate what you need. 
if i see other wallet or aggregator demo site has this kind of hardcoded address, especially as a send-to address, i would find it very fishy.
cc @nateReiners

spencerstock:
> @spencerstock why not running a new dapp dedicated to do so and automate what you need. if i see other wallet or aggregator demo site has this kind of hardcoded address, especially as a send-to address, i would find it very fishy. cc @nateReiners

@bangtoven Would you be alright with `0x0000000000000000000000000000000000000000` the universal black hole address?

apps/testapp/src/components/RpcMethods/shortcut/sendShortcuts.ts
```diff
@@ -6,7 +6,7 @@ const ethSendTransactionShortcuts: ShortcutType[] = [
     key: 'Example Tx',
     data: {
       from: ADDR_TO_FILL,
-      to: ADDR_TO_FILL,
+      to: '0x0000000000000000000000000000000000000000',
```
eric-ships, line 9: Could be good to have a comment along the lines of "Because users are blocked on smart wallets from having a `to` address of itself for security and safety reasons, using the universal blackhole address instead."

eric-ships (approved):
Looks good

spencerstock (approved):

PR #1299, opened 2024-05-21
Add alert when popup is blocked

arjun-dureja:
### _Summary_

- While it's unlikely, the popup may get blocked on some browsers due to strict popup blockers. This PR adds an alert to inform the user to disable their popup blocker when the popup gets blocked.

- Manually and unit tests

How to test:
1. Comment out [this line](https://github.com/coinbase/coinbase-wallet-sdk/blob/master/packages/wallet-sdk/src/sign/scw/SCWSigner.ts#L83)
2. Attempt to sign a transaction using the test app

https://github.com/coinbase/coinbase-wallet-sdk/assets/16995513/7b6584a3-8bef-4226-a43f-367124e960ea

### DIFF
--- a/packages/wallet-sdk/src/core/communicator/util.test.ts
+++ b/packages/wallet-sdk/src/core/communicator/util.test.ts
@@ -59,7 +59,9 @@ describe('Communicator util', () => {
 
       expect(() => openPopup(url)).toThrowError('Pop up window failed to open');
       expect(window.alert).toHaveBeenCalledTimes(1);
-      expect(window.alert).toHaveBeenCalledWith('Please disable your pop-up blocker and try again');
+      expect(window.alert).toHaveBeenCalledWith(
+        'Smart wallet pop up failed, please disable your pop up blocker and try again.'
+      );
     });
   });
 
--- a/packages/wallet-sdk/src/core/communicator/util.test.ts
+++ b/packages/wallet-sdk/src/core/communicator/util.test.ts
@@ -1,0 +1,91 @@
+import { closePopup, openPopup } from './util';
+
+describe('Communicator util', () => {
+  afterEach(() => {
+    jest.restoreAllMocks();
+  });
+
+  describe('openPopup', () => {
+    it('opens a popup window', () => {
+      const url = new URL('https://keys.coinbase.com');
+      const mockFocus = jest.fn();
+
+      const windowSpy = jest.spyOn(window, 'open').mockImplementation(() => {
+        return {
+          focus: mockFocus,
+        } as any;
+      });
+
+      openPopup(url);
+
+      expect(windowSpy).toHaveBeenCalledTimes(1);
+      expect(windowSpy).toHaveBeenCalledWith(url, 'Smart Wallet', expect.any(String));
+      expect(mockFocus).toHaveBeenCalled();
+    });
+
+    it('uses the correct dimensions and location for the popup window', () => {
+      const url = new URL('https://keys.coinbase.com');
+
+      const mockScreenXY = 100;
+      const mockScreenInnerWidthHeight = 50;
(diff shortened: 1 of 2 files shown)


packages/wallet-sdk/src/core/communicator/util.ts
```diff
@@ -14,10 +14,15 @@ export function openPopup(url: URL): Window {
...
     `width=${POPUP_WIDTH}, height=${POPUP_HEIGHT}, left=${left}, top=${top}`
   );
-  popup?.focus();
-  if (!popup) {
+
+  if (popup) {
+    popup.focus();
+  } else {
+    // The popup may fail to open if the user has a pop-up blocker enabled
+    alert('Please disable your pop-up blocker and try again');
```
bangtoven, line 22: it's not a scope of this pr, but would it be possible to listen to user event on alert closing and retry popup open? @arjun-dureja

arjun-dureja: I think it's technically possible, but may cause problems because the user would need to close the alert before they can go to their settings to disable popup blocking. So if we try again immediately after closing, it'll just result in another alert/failure

packages/wallet-sdk/src/core/communicator/util.test.ts
```diff
@@ -0,0 +1,91 @@
...
+      const expectedWidth = 420;
+      const expectedHeight = 540;
+      const expectedLeft = (mockScreenInnerWidthHeight - expectedWidth) / 2 + mockScreenXY;
+      const expectedTop = (mockScreenInnerWidthHeight - expectedHeight) / 2 + mockScreenXY;
+      expect(windowSpy).toHaveBeenCalledWith(
+        url,
+        'Smart Wallet',
+        `width=${expectedWidth}, height=${expectedHeight}, left=${expectedLeft}, top=${expectedTop}`
+      );
+    });
```
bangtoven, line 53: image

packages/wallet-sdk/src/core/communicator/util.ts
bangtoven: i mean, i think we can make the popup open call directly and immediately to get triggered via a specific user action, so theoretically browsers should not block it, no?

packages/wallet-sdk/src/core/communicator/util.test.ts
arjun-dureja: The test ensures that the top and left values are calculated correctly. If the calculation were to change, this test would fail and help us spot a bug

bangtoven: image
@arjun-dureja

packages/wallet-sdk/src/core/communicator/util.test.ts
```diff
@@ -0,0 +1,91 @@
...
+          focus: jest.fn(),
+        } as any;
+      });
+
+      openPopup(url);
+
+      const expectedWidth = 420;
+      const expectedHeight = 540;
+      const expectedLeft = (mockScreenInnerWidthHeight - expectedWidth) / 2 + mockScreenXY;
+      const expectedTop = (mockScreenInnerWidthHeight - expectedHeight) / 2 + mockScreenXY;
```
bangtoven, line 47: image
this is just repetition of the same code

packages/wallet-sdk/src/core/communicator/util.ts
bangtoven: web noob question.
can we use `confirm()` and retry opening when it returns true?

bangtoven (approved):

bangtoven:
thanks a lot @arjun-dureja for exploring these options 🙏 🙏 🙏 
let's figure out how we can offer more refined experience.

PR #1301, opened 2024-05-22
Fix popup not opening on iPad

arjun-dureja:
### _Summary_

- There's an issue on iPads where the object returned from `window.open` stays defined even after the window/tab is closed. This causes the popup to not be opened when one was opened in the past due to this line of code:
https://github.com/coinbase/coinbase-wallet-sdk/blob/53f37742c40ff5c20f6c652611b322c318332328/packages/wallet-sdk/src/core/communicator/Communicator.ts#L82-L83

- To solve this, I added a check for `popup.closed` on the same line, so that the popup is opened again when it's closed

- Manually and unit tests
- Also verified that it's still working as expected on other devices

| Before  | After |
| ------------- | ------------- |
| <video src="https://github.com/coinbase/coinbase-wallet-sdk/assets/16995513/f1043bb0-79c9-4c22-b45c-1d846c6f5db7">  | <video src="https://github.com/coinbase/coinbase-wallet-sdk/assets/16995513/380bcbb4-96ad-43bb-8486-0befef5eb463">|

### DIFF
--- a/packages/wallet-sdk/src/core/communicator/Communicator.test.ts
+++ b/packages/wallet-sdk/src/core/communicator/Communicator.test.ts
@@ -53,6 +53,7 @@ describe('Communicator', () => {
     mockPopup = {
       postMessage: jest.fn(),
       close: jest.fn(),
+      closed: false,
     } as unknown as Window;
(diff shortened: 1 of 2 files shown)


bangtoven (approved):
🔥

PR #1309, opened 2024-05-28
fix default appChainIds

nateReiners:


bangtoven:
what problem does it solve?

bangtoven:
re-open if needed @nateReiners

PR #1310, opened 2024-05-30
make eth_chainId comply with spec

nateReiners:
### _Summary_
Make `eth_chainId` match the spec. It should return a hex string.

https://ethereum.org/en/developers/docs/apis/json-rpc/#eth_chainid


fan-zhang-sv (commented):
nice!

fan-zhang-sv (approved):

PR #1315, opened 2024-06-06
Re-focus the popup between requests

arjun-dureja:
### _Summary_

- It's possible for a user to leave the popup open between requests, especially after onboarding when the popup is not closed programmatically. Because of this, when a new request comes through, the popup doesn't re-focus itself and updates in the background. This leads to a bad UX since the user can't see the popup and has to manually re open it.
  - This PR simply calls `popup.focus()` every time a new request comes to an already open popup. This ensures that the popup doesn't stay hidden in the background when a user performs an action on a dapp

- Manually and unit tests

### **BEFORE:**
https://github.com/coinbase/coinbase-wallet-sdk/assets/16995513/47df2600-d3e0-4864-bd5c-c3918473838a

### **AFTER:**

https://github.com/coinbase/coinbase-wallet-sdk/assets/16995513/7e738c17-869f-43aa-8088-083ffe2cf1c7

### DIFF
--- a/packages/wallet-sdk/src/core/communicator/Communicator.test.ts
+++ b/packages/wallet-sdk/src/core/communicator/Communicator.test.ts
@@ -44,7 +44,10 @@ const removeEventListenerSpy = jest.spyOn(window, 'removeEventListener');
 describe('Communicator', () => {
   let urlOrigin: string;
   let communicator: Communicator;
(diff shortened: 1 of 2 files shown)


nateReiners (approved):
lgtm, good idea! 👍

PR #1321, opened 2024-06-21
Don't throw when grabbing injected provider

fan-zhang-sv:
### _Summary_

window.top may throw DOMException, we should not let this error blocking us from continuing

### DIFF
--- a/packages/wallet-sdk/src/util/provider.ts
+++ b/packages/wallet-sdk/src/util/provider.ts
@@ -47,24 +47,27 @@ export function getCoinbaseInjectedProvider({
   metadata,
   preference,
 }: Readonly<ConstructorOptions>): ProviderInterface | undefined {
-  const window = globalThis as CBWindow;
+  try {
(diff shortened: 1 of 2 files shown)


nateReiners (approved):
lgtm

cb-heimdall:
Review Error for nateReiners @ 2024-06-21 20:31:14 UTC

packages/wallet-sdk/src/util/provider.ts
```diff
@@ -61,10 +61,13 @@ export function getCoinbaseInjectedProvider({
     }
   }
 
-  const ethereum = window.ethereum ?? window.top?.ethereum;
```
bangtoven, line 64: how about creating a dedicated function or something to do just this and catch error?
the if statement below doesn’t throw error anyway.

nateReiners (approved):

bangtoven (commented):
@fan-zhang-sv

PR #1322, opened 2024-06-21
Helper to Get Provider from Top Frame

fan-zhang-sv:

### DIFF
--- a/packages/wallet-sdk/src/util/provider.ts
+++ b/packages/wallet-sdk/src/util/provider.ts
@@ -43,7 +43,7 @@ export function getCoinbaseInjectedSigner(): Signer | undefined {
   return window.coinbaseWalletSigner;
 }
 
-function getCipherWebViewInjectedProvider(): CBInjectedProvider | undefined {
+function getInjectedEthereum(): CBInjectedProvider | undefined {
   const window = globalThis as CBWindow;
   try {
     return window.ethereum ?? window.top?.ethereum;
@@ -71,7 +71,7 @@ export function getCoinbaseInjectedProvider({
     }
   }
 
-  const ethereum = getCipherWebViewInjectedProvider();
+  const ethereum = getInjectedEthereum();
   if (ethereum?.isCoinbaseBrowser) {
     return ethereum;
   }
--- a/packages/wallet-sdk/src/util/provider.ts
+++ b/packages/wallet-sdk/src/util/provider.ts
@@ -43,6 +43,16 @@ export function getCoinbaseInjectedSigner(): Signer | undefined {
   return window.coinbaseWalletSigner;
 }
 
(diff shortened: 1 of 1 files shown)


packages/wallet-sdk/src/util/provider.ts
```diff
@@ -43,6 +43,16 @@ export function getCoinbaseInjectedSigner(): Signer | undefined {
   return window.coinbaseWalletSigner;
 }
 
+function getTopFrameInjectedProvider(): CBInjectedProvider | undefined {
```
bangtoven, line 46: how about calling it getInjectedProvider or something 
and encapsulate all window.ethereum logic?

bangtoven (commented):
@fan-zhang-sv

packages/wallet-sdk/src/util/provider.ts
fan-zhang-sv: pushed, how does `getCipherWebViewInjectedProvider` sound?

bangtoven: to be it seems to be just simple helper to retrieve injected provider from any of the window properties, either window or window.top
and cipher web view check happens the next line

fan-zhang-sv: agree, `getInjectedEthereum`?

bangtoven (approved):

packages/wallet-sdk/src/util/provider.ts
```diff
@@ -43,6 +43,16 @@ export function getCoinbaseInjectedSigner(): Signer | undefined {
...
 }
 
+function getInjectedEthereum(): CBInjectedProvider | undefined {
+  const window = globalThis as CBWindow;
+  try {
+    return window.ethereum ?? window.top?.ethereum;
+    // eslint-disable-next-line no-empty
+  } catch {}
+
+  return undefined;
```
bangtoven, line 53: how about this?
```suggestion
  try {
    const window = globalThis as CBWindow;
    return window.ethereum ?? window.top?.ethereum;
  } catch {
    return undefined;
  }
```

packages/wallet-sdk/src/util/provider.ts
```diff
@@ -43,6 +43,16 @@ export function getCoinbaseInjectedSigner(): Signer | undefined {
...
+}
+
 export function getCoinbaseInjectedProvider({
   metadata,
   preference,
```
bangtoven, line 58: we might remove the line 50, 
and do this instead for line 56, `const extension =(globalThis as CBWindow).coinbaseWalletExtension;`

bangtoven: or, seems bit overkill, but for the sake of consistency, we might create a dedicated helper method for this maybe?

PR #1336, opened 2024-06-28
replace sha.js and keccak with smaller and more secure @noble/hashes

talentlessguy:
### _Summary_

This PR replaces keccak and sha.js with a smaller, audited and more secure `@noble/hashes`, thus removing 11 deps.

Before (15 deps):

image

After (4 deps):

image

All tests passed.

### DIFF
--- a/packages/wallet-sdk/package.json
+++ b/packages/wallet-sdk/package.json
@@ -28,11 +28,10 @@
     "lint": "eslint . --ext .ts,.tsx --fix"
   },
   "dependencies": {
+    "@noble/hashes": "^1.4.0",
     "buffer": "^6.0.3",
     "clsx": "^1.2.1",
     "eventemitter3": "^5.0.1",
-    "js-sha256": "^0.11.0",
-    "keccak": "^3.0.3",
     "preact": "^10.16.0"
   },
   "devDependencies": {
--- a/packages/wallet-sdk/package.json
+++ b/packages/wallet-sdk/package.json
@@ -64,6 +64,5 @@
     "tsc-alias": "^1.8.8",
     "tslib": "^2.6.0",
     "typescript": "^5.1.6"
-  },
-  "packageManager": "yarn@1.22.22+sha1.ac34549e6aa8e7ead463a7407e1c7390f61a6610"
+  }
 }
--- a/packages/wallet-sdk/package.json
+++ b/packages/wallet-sdk/package.json
@@ -31,9 +31,9 @@
     "buffer": "^6.0.3",
     "clsx": "^1.2.1",
     "eventemitter3": "^5.0.1",
+    "js-sha256": "^0.11.0",
     "keccak": "^3.0.3",
-    "preact": "^10.16.0",
-    "sha.js": "^2.4.11"
+    "preact": "^10.16.0"
   },
   "devDependencies": {
     "@babel/core": "^7.22.9",
@@ -46,7 +46,6 @@
     "@testing-library/preact": "^2.0.1",
     "@types/jest": "^27.5.2",
     "@types/node": "^14.18.54",
-    "@types/sha.js": "^2.4.1",
     "@typescript-eslint/eslint-plugin": "^6.2.0",
     "@typescript-eslint/parser": "^6.2.0",
     "babel-jest": "^27.5.1",
--- a/packages/wallet-sdk/package.json
+++ b/packages/wallet-sdk/package.json
@@ -65,5 +65,6 @@
     "tsc-alias": "^1.8.8",
     "tslib": "^2.6.0",
     "typescript": "^5.1.6"
-  }
+  },
+  "packageManager": "yarn@1.22.22+sha1.ac34549e6aa8e7ead463a7407e1c7390f61a6610"
 }
--- a/packages/wallet-sdk/package.json
+++ b/packages/wallet-sdk/package.json
@@ -29,7 +29,6 @@
   },
   "dependencies": {
     "@noble/hashes": "^1.4.0",
-    "buffer": "^6.0.3",
     "clsx": "^1.2.1",
     "eventemitter3": "^5.0.1",
     "preact": "^10.16.0"
--- a/packages/wallet-sdk/src/sign/walletlink/relay/type/WalletLinkSession.ts
+++ b/packages/wallet-sdk/src/sign/walletlink/relay/type/WalletLinkSession.ts
@@ -1,6 +1,6 @@
 // Copyright (c) 2018-2023 Coinbase, Inc. <https://www.coinbase.com/>
 
-import { sha256 } from 'js-sha256';
+import { sha256 } from '@noble/hashes/sha256';
 
 import { randomBytesHex } from ':core/type/util';
 import { ScopedLocalStorage } from ':util/ScopedLocalStorage';
@@ -21,10 +21,9 @@ export class WalletLinkSession {
     this._id = id || randomBytesHex(16);
     this._secret = secret || randomBytesHex(32);
 
-    this._key = sha256
-      .create()
-      .update(`${this._id}, ${this._secret} WalletLink`) // ensure old sessions stay connected
-      .hex();
+    const hash = sha256.create().update(`${this._id}, ${this._secret} WalletLink`).digest();
+
+    this._key = Array.from(hash, (byte) => byte.toString(16).padStart(2, '0')).join('');
 
     this._linked = !!linked;
   }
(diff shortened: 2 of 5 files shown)


talentlessguy:
Another option is using `@noble/curves` which are audited and used by most of Ethereum cryptography, but it would pull the code for the whole package, not just sha256

bangtoven:
@talentlessguy thanks a lot for the improvement! @noble/curves seems nice. should we consider having that instead? what about using viem? that might be already imported by many of the projects. hdyt?

paulmillr:
>but it would pull the code for the whole package, not just sha256

That's false. sha256-only build does not include other hashes or primitives. It's also very friendly to tree-shaking even when no tools support it. Usage:

```ts
import { sha256 } from '@noble/hashes/sha256';
```

bangtoven:
hey @paulmillr thanks for chiming in and clarifying!
> viem is using noble-hashes and noble-curves underneath.
that sounds great. there are some legacy code that we relies on old dependencies, and viem might be a preferable option. good to know that we can use noble through it.

talentlessguy:
I will replace it with noble curves because there's no point in pulling viem when only a single cryptography function is used, IMO

I hope that's alright

talentlessguy:
> > but it would pull the code for the whole package, not just sha256
> 
> That's false. sha256-only build does not include other hashes or primitives. It's also very friendly to tree-shaking even when no tools support it.

Sorry for bad phrasing, I didn't mean the app pulling it by npm pulling the whole source instead of individual files (only ones for sha256)

paulmillr:
> npm pulling the whole source instead of individual files

Why is that a problem? Are you paying for your network traffic so that 1MB would cost much more than 0.1MB package?

talentlessguy:
> > npm pulling the whole source instead of individual files
> 
> Why is that a problem? Are you paying for your network traffic so that 1MB would cost much more than 0.1MB package?

it's a problem not related to noble hashes, just how npm is designed. it massively slows down package installation and wastes space on disk and decreases it's lifespan. It's okay if a few deps are heavy but when everybody does this, it becomes a problem. But this is if off-topic.

On-topic - I replaced keccak with noble/hashes/sha3. Now cb SDK has less deps and better security

talentlessguy:
@bangtoven if there's any additional changes required, please let me know. All tests pass, formatting is correct.

packages/wallet-sdk/src/sign/walletlink/relay/type/WalletLinkSession.ts
```diff
@@ -21,10 +21,9 @@ export class WalletLinkSession {
     this._id = id || randomBytesHex(16);
     this._secret = secret || randomBytesHex(32);
 
-    this._key = sha256
-      .create()
-      .update(`${this._id}, ${this._secret} WalletLink`) // ensure old sessions stay connected
-      .hex();
+    const hash = sha256.create().update(`${this._id}, ${this._secret} WalletLink`).digest();
+
+    this._key = Array.from(hash, (byte) => byte.toString(16).padStart(2, '0')).join('');
```
paulmillr, line 26: `import { bytesToHex } from '@noble/hashes/utils'` is faster

paulmillr:
> it massively slows down package installation and wastes space on disk and decreases it's lifespan

10 avg dependencies slow down stuff more than one bigger no-deps module. Also, there is no slow down after first install: it would be cached. I could have made hashes.tgz like 4x smaller if I removed source maps, double types, etc. But it's all there for developer convenience.

Everyone is using like 500 deps everywhere. No one audits them, no one knows what's inside. They massively slow down everything. This is the big problem.

paulmillr:
I'm asking to focus on relevant metrics (traffic consumed by users when they visit website) instead of vanity metrics (traffic consumed by developers when developing):

```
size ❯ npx esbuild --bundle input-noble.js --outfile=noble-sha3.js --global-name=out

  noble-sha3.js  10.6kb

size ❯ npx esbuild --bundle input.js --outfile=jssha3.js --global-name=out  

  jssha3.js  26.1kb
```

It's important to care about end-user.

packages/wallet-sdk/src/sign/walletlink/relay/type/WalletLinkSession.ts
talentlessguy: thanks for the suggestion, will use that! didn't know noble hashes already include this util

talentlessguy:
> I'm asking to focus on relevant metrics (traffic consumed by users when they visit website) instead of vanity metrics (traffic consumed by developers when developing):
> 
> ```
> size ❯ npx esbuild --bundle input-noble.js --outfile=noble-sha3.js --global-name=out
> 
>   noble-sha3.js  10.6kb
> 
> size ❯ npx esbuild --bundle input.js --outfile=jssha3.js --global-name=out  
> 
>   jssha3.js  26.1kb
> ```
> 
> It's important to care about end-user.

yeah I agree. dep count is not the only important metric. deps should also he tree-shakeable and as small as possible. It's super awesome that noble hashes is both.

in my perspective, both DX and UX should be good, nobody should be left out. And oftentimes bad DX results in poor UX.

packages/wallet-sdk/src/sign/walletlink/relay/type/WalletLinkSession.ts
talentlessguy: fixed

bangtoven:
THANK YOU @talentlessguy (isn't your username so ironic 😉) and @paulmillr for thoughtful discussion to provide better experience for both developers and end-users!
reducing 15 deps to 4 deps is amazing!! 

i have to admit, this repo didn't have proper maintenance for quite long time. and we'd want to clean it up to follow more modern and standard conventions to better serve the community.
this is yet another legacy setup that we'd want to get rid of:
https://github.com/coinbase/coinbase-wallet-sdk/tree/master/packages/wallet-sdk/src/vendor-js/eth-eip712-util
i assumed that viem is popular enough that many of the dapps are already importing and could be resolved without adding extra dependency.
how do you think?

paulmillr:
>i assumed that viem is popular enough that many of the dapps are already importing and could be resolved without adding extra dependency.

It is popular and they import it but it doesn't mean apps would use same viem as you use.

Which means, there may be 2 viems. One from coinbase and one from the app itself. Definitely the worst-case scenario.

bangtoven:
@paulmillr yeah good point. and i just realized that viem is actually bigger than i imagined https://bundlephobia.com/package/viem@2.17.3

packages/wallet-sdk/src/vendor-js/eth-eip712-util/index.js
```diff
@@ -1,5 +1,6 @@
-const util = require('./util')
-const abi = require('./abi')
+/* eslint-disable @typescript-eslint/no-var-requires */
+const util = require('./util');
+const abi = require('./abi');
```
bangtoven, line 3: seems like all these changes on this file are by linter/prettier, right?

talentlessguy: yeah I've ran prettier script to ensure fmt is right. I can remove it from diff if needed

jxom:
> @paulmillr yeah good point. and i just realized that viem is actually bigger than i imagined https://bundlephobia.com/package/viem@2.17.3

Bundlephobia is a very bad source for measuring bundle size. It doesn't account for tree-shaking. Minimal surface size for Viem is ~5kB.

bangtoven:
hey @jxom ! 👋 thanks for chiming in. 
that's great to know that! i'd love to replace all of our legacy cryptographic implementations with viem. will take a stab at it. thanks!

packages/wallet-sdk/src/vendor-js/eth-eip712-util/index.js
bangtoven: @talentlessguy gotcha. yeah, can we try minimizing the diff? i honestly don't care much about those .js files, bc we'd want to deprecate them soon anyway. 
unless it needs changes (i think i saw one of those .js files was using the old dep) let's just leave it that way.

packages/wallet-sdk/package.json
```diff
@@ -67,5 +64,6 @@
     "tsc-alias": "^1.8.8",
     "tslib": "^2.6.0",
     "typescript": "^5.1.6"
-  }
+  },
+  "packageManager": "yarn@1.22.22+sha1.ac34549e6aa8e7ead463a7407e1c7390f61a6610"
```
fan-zhang-sv, line 68: hi hi, this line is added possibly because the yarn you used was coming from Node.js `corepack`, could you remove this line and try use a globally installed yarn to run yarn command and see if this line persists?

talentlessguy: yeah sure, I keep forgetting to remove it cuz it gets added every time I run yarn

bangtoven (approved):
This is amazing contribution, not just for this repo but for whole crypto community. Thank you @talentlessguy 
image

talentlessguy:
huge thx for merging, is there an ETA on when this gets published?

PR #1339, opened 2024-07-08
cleanup prep for ext signer

nateReiners:


packages/wallet-sdk/src/util/provider.ts
```diff
@@ -29,11 +29,15 @@ export async function fetchRPCRequest(request: RequestArguments, chain?: Chain)
 export interface CBWindow {
```
bangtoven, line 29: thanks for separating this out.
can we just get rid of all of this? there's no point for us to define `CBWindow` `CBMobileInjectedProvider` nor `CBExtensionInjectedProvider`.
i think we can remove this entire file.

PR #1343, opened 2024-07-12
#trivial simplify handshake return value

bangtoven:
### _Summary_

we don't need return value from handshake. it's not being used on provider. we only care if it was successful or not. Promise<void> should be enough

 unit tests

### DIFF
--- a/packages/wallet-sdk/src/sign/extension/ExtensionSigner.ts
+++ b/packages/wallet-sdk/src/sign/extension/ExtensionSigner.ts
@@ -48,11 +48,9 @@ export class ExtensionSigner implements Signer {
   }
 
   async handshake() {
-    const accounts = (await this.request({
+    await this.request({
       method: 'eth_requestAccounts',
-    })) as AddressString[];
-    this.updateListener.onAccountsUpdate(accounts);
-    return accounts;
+    });
   }
 
   async request(request: RequestArguments) {
--- a/packages/wallet-sdk/src/sign/interface.ts
+++ b/packages/wallet-sdk/src/sign/interface.ts
@@ -4,7 +4,7 @@ import { AddressString, Chain } from ':core/type';
 export interface Signer {
   readonly accounts: AddressString[];
   readonly chain: Chain;
-  handshake(): Promise<AddressString[]>;
+  handshake(): Promise<void>;
   request(request: RequestArguments): Promise<unknown>;
   disconnect: () => Promise<void>;
 }
--- a/packages/wallet-sdk/src/sign/walletlink/WalletLinkSigner.ts
+++ b/packages/wallet-sdk/src/sign/walletlink/WalletLinkSigner.ts
@@ -100,10 +100,6 @@ export class WalletLinkSigner implements Signer {
     return { id, secret };
   }
 
-  async handshake() {
-    return this.request({ method: 'eth_requestAccounts' });
-  }
-
   get selectedAddress(): AddressString | undefined {
     return this._addresses[0] || undefined;
   }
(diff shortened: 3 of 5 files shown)


packages/wallet-sdk/src/sign/walletlink/WalletLinkSigner.ts
```diff
@@ -492,13 +485,9 @@ export class WalletLinkSigner implements Signer {
     return ensureIntNumber(chainId);
   }
 
-  private async _eth_requestAccounts(): Promise<JSONRPCResponse> {
+  async handshake() {
```
bangtoven, line 488: this logic below is what we want from WalletLinkSigner for its handshake process.
instead of invoking it indirectly via `return this.request({ method: 'eth_requestAccounts' });`, we can rename this function to be our `handshake` definition.
image

packages/wallet-sdk/src/sign/extension/ExtensionSigner.ts
```diff
@@ -48,11 +48,9 @@ export class ExtensionSigner implements Signer {
   }
 
   async handshake() {
-    const accounts = (await this.request({
+    await this.request({
       method: 'eth_requestAccounts',
-    })) as AddressString[];
-    this.updateListener.onAccountsUpdate(accounts);
```
bangtoven, line 54: redundant. extension should emit this anyway
image

packages/wallet-sdk/src/sign/interface.ts
```diff
@@ -4,7 +4,7 @@ import { AddressString, Chain } from ':core/type';
 export interface Signer {
   readonly accounts: AddressString[];
   readonly chain: Chain;
-  handshake(): Promise<AddressString[]>;
+  handshake(): Promise<void>;
```
bangtoven, line 7: we don't need return value from handshake. it's not being used on provider. we only care if it was successful or not. Promise should be enough

packages/wallet-sdk/src/sign/walletlink/WalletLinkSigner.ts
```diff
@@ -303,8 +299,10 @@ export class WalletLinkSigner implements Signer {
     const params = request.params || [];
 
     switch (method) {
-      case 'eth_requestAccounts':
-        return this._eth_requestAccounts();
+      case 'eth_requestAccounts': {
+        await this.handshake();
+        return { jsonrpc: '2.0', id: 0, result: this._addresses };
```
bangtoven, line 304: just following the old code adding this nonsense `jsonrpc: '2.0', id: 0,` 
image
we don't use them anywhere, so it's probably okay to deprecate. will revisit this when we refactor WalletLink implementation

packages/wallet-sdk/src/sign/extension/ExtensionSigner.ts
nateReiners: don't we need to set the accounts on the signer here so the provider can return them on L66 here
Screenshot 2024-07-12 at 4 00 17 PM

nateReiners: `this._accounts = accounts` in the handshake, and then add getters for `accounts` and `chain`

fan-zhang-sv (approved):
lgtm

packages/wallet-sdk/src/sign/extension/ExtensionSigner.ts
bangtoven: that return value isn't being used at all.
and we have getter.
@nateReiners

nateReiners: oh I see :P then... should we delete L66

PR #1345, opened 2024-07-16
Allow consumer to pass in storage object

vishnumad:
### _Summary_

* Allow consumer dapp to pass in a custom `BaseStorage` object, which will be used by the SDK internally
* Creates `browserStorageAdapter` and `mmkvStorageAdapter` helper functions which can be used to transform `localStorage` or React Native MMKV storage objects into a `BaseStorage`
* Renames `ScopedLocalStorage` to `ScopedStorage`

Tested manually + unit tests

### DIFF
--- a/packages/wallet-sdk/src/CoinbaseWalletProvider.ts
+++ b/packages/wallet-sdk/src/CoinbaseWalletProvider.ts
@@ -18,7 +18,7 @@ import { SignerType } from ':core/message';
 import { determineMethodCategory } from ':core/provider/method';
 import { hexStringFromNumber } from ':core/type/util';
 import type { BaseStorage } from ':util/BaseStorage';
-import { ScopedLocalStorage } from ':util/ScopedLocalStorage';
+import { ScopedStorage } from ':util/ScopedStorage';
 
 export class CoinbaseWalletProvider extends EventEmitter implements ProviderInterface {
   private readonly metadata: AppMetadata;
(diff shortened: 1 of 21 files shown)


packages/wallet-sdk/src/CoinbaseWalletProvider.ts
```diff
@@ -17,22 +17,29 @@ import { Communicator } from ':core/communicator/Communicator';
...
   private readonly metadata: AppMetadata;
   private readonly preference: Preference;
   private readonly communicator: Communicator;
+  private readonly baseStorage?: BaseStorage;
 
   private signer: Signer | null;
 
-  constructor({ metadata, preference: { keysUrl, ...preference } }: Readonly<ConstructorOptions>) {
+  constructor({
+    baseStorage,
```
bangtoven, line 32: so if i understand this correctly,
the branching, e.g. `const baseStorage = isMobile? mobileStorage : webStorage` would happen outside, which means the consumer of this class will be responsible for it, right?
@vishnumad

fan-zhang-sv (approved):
LGTM! the storage class might need a clean up due to historical reason but this PR is pretty good, let's merge to unblock the goal of getting MVP out asap!

packages/wallet-sdk/src/CoinbaseWalletProvider.ts
vishnumad: @bangtoven, going to merge this to unblock mobile sdk progress. Let's revisit encapsulating this logic in the sdk as a followup.

PR #1349, opened 2024-07-22
remove react as dev dep for now

fan-zhang-sv:
### _Summary_

react dev dep is causing issue in our current development pipeline, improving this may require restructure the whole project, removing it for now as its just for one test file only.


vishnumad (approved):

bangtoven:
@felix should we merge this in?

bangtoven:
@fan-zhang-sv should we merge this in?

PR #1402, opened 2024-09-23
add support for any config parameter

cb-jake:
### _Summary_

This PR extends the preferences object to support any config value. This PR also exposes new methods to get the provider or an new sdk object. 

Manually. In a follow up PR there will be an update to the test app to use the new create SDK object

### DIFF
--- a/packages/wallet-sdk/src/CoinbaseWalletProvider.ts
+++ b/packages/wallet-sdk/src/CoinbaseWalletProvider.ts
@@ -30,7 +30,9 @@ export class CoinbaseWalletProvider extends ProviderEventEmitter implements Prov
     this.communicator = new Communicator(keysUrl, metadata);
 
     const signerType = loadSignerType();
-    if (signerType) this.signer = this.initSigner(signerType);
+    if (signerType) {
+      this.signer = this.initSigner(signerType);
+    }
   }
 
   public async request(args: RequestArguments): Promise<unknown> {
--- a/packages/wallet-sdk/src/CoinbaseWalletProvider.ts
+++ b/packages/wallet-sdk/src/CoinbaseWalletProvider.ts
@@ -12,6 +12,7 @@ import { Signer } from './sign/interface';
 import { createSigner, fetchSignerType, loadSignerType, storeSignerType } from './sign/util';
 import { checkErrorForInvalidRequestArgs } from './util/provider';
 import { Communicator } from ':core/communicator/Communicator';
+import { CB_KEYS_URL } from ':core/constants';
 import { SignerType } from ':core/message';
 import { ScopedLocalStorage } from ':core/storage/ScopedLocalStorage';
 import { hexStringFromNumber } from ':core/type/util';
@@ -27,7 +28,9 @@ export class CoinbaseWalletProvider extends ProviderEventEmitter implements Prov
     super();
     this.metadata = metadata;
     this.preference = preference;
-    this.communicator = new Communicator(keysUrl, metadata);
+
+    const url = typeof keysUrl === 'string' ? keysUrl : CB_KEYS_URL;
+    this.communicator = new Communicator(url, metadata);
 
     const signerType = loadSignerType();
     if (signerType) {
--- a/packages/wallet-sdk/src/CoinbaseWalletProvider.ts
+++ b/packages/wallet-sdk/src/CoinbaseWalletProvider.ts
@@ -12,7 +12,6 @@ import { Signer } from './sign/interface';
 import { createSigner, fetchSignerType, loadSignerType, storeSignerType } from './sign/util';
 import { checkErrorForInvalidRequestArgs } from './util/provider';
 import { Communicator } from ':core/communicator/Communicator';
-import { CB_KEYS_URL } from ':core/constants';
 import { SignerType } from ':core/message';
 import { ScopedLocalStorage } from ':core/storage/ScopedLocalStorage';
 import { hexStringFromNumber } from ':core/type/util';
@@ -28,9 +27,7 @@ export class CoinbaseWalletProvider extends ProviderEventEmitter implements Prov
     super();
     this.metadata = metadata;
     this.preference = preference;
-
-    const url = typeof keysUrl === 'string' ? keysUrl : CB_KEYS_URL;
-    this.communicator = new Communicator(url, metadata);
+    this.communicator = new Communicator(keysUrl, metadata);
 
     const signerType = loadSignerType();
     if (signerType) {
--- a/packages/wallet-sdk/src/core/provider/interface.ts
+++ b/packages/wallet-sdk/src/core/provider/interface.ts
@@ -51,11 +51,10 @@ type OnrampPrefillOptions = {
 };
 
 export type Preference = {
-  options: 'all' | 'smartWalletOnly' | 'eoaOnly';
   /**
-   * @deprecated this is only for internal use
+   * @param options
    */
-  keysUrl?: string;
+  options: 'all' | 'smartWalletOnly' | 'eoaOnly';
   /**
    * @param postOnboardingAction
    * @type {PostOnboardingAction}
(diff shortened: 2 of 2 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

packages/wallet-sdk/src/core/provider/interface.ts
```diff
@@ -42,10 +42,47 @@ export interface AppMetadata {
...
+  amount: string;
+  chainId: number;
+};
+
+export type Preference = {
   options: 'all' | 'smartWalletOnly' | 'eoaOnly';
+  /**
+   * @deprecated this is only for internal use
+   */
   keysUrl?: string;
```
fan-zhang-sv, line 58: since this is only used by us, can we remove it for now

if anything, we will always catch the problem before we roll out as this is used internally i think?

fan-zhang-sv (commented):
looking good, left a comment!

packages/wallet-sdk/src/core/provider/interface.ts
```diff
@@ -42,10 +42,47 @@ export interface AppMetadata {
...
-}
+  /**
+   * @param postOnboardingAction
+   * @type {PostOnboardingAction}
+   * @description Recommends the action to be taken after the onboarding process to engage users with various flows. Possible values:
+   * - `none`: No action is recommended post-onboarding. (Default experience)
+   * - `onramp`: Recommends initiating the onramp flow, allowing users to prefill their account with a specified asset, chain, and amount.
```
spencerstock, line 63: > allowing users to prefill their account with a specified asset, chain, and amount.

"allowing users to prefill their account with an optional <OnrampPrefillOptions>"

packages/wallet-sdk/src/core/provider/interface.ts
cb-jake: Might be good to list it as deprecated for a few versions just in case someone externally is using it

packages/wallet-sdk/src/core/provider/interface.ts
```diff
@@ -42,10 +42,47 @@ export interface AppMetadata {
...
   options: 'all' | 'smartWalletOnly' | 'eoaOnly';
+  /**
+   * @deprecated this is only for internal use
+   */
   keysUrl?: string;
-}
+  /**
+   * @param postOnboardingAction
+   * @type {PostOnboardingAction}
+   * @description Recommends the action to be taken after the onboarding process to engage users with various flows. Possible values:
```
spencerstock, line 61: > Recommends the action to be taken after the onboarding process to engage users with various flows. Possible values:

I think we could add some clarity here
"Displays CTAs to the user based on the preference of the dapp. These CTAs are part of prebuilt UI components that are available to the CB Smart Wallet"

packages/wallet-sdk/src/core/provider/interface.ts
```diff
@@ -42,10 +42,47 @@ export interface AppMetadata {
...
+   */
+  onrampPrefillOptions?: OnrampPrefillOptions;
+  /**
+   * @param attributionDataSuffix
+   * @type {Hex}
+   * @description Data suffix to be appended to the initCode or executeBatch calldata
+   * Expects a 4 byte hex string
+   */
+  attributionDataSuffix?: string;
+} & Record<string, unknown>;
```
spencerstock, line 84: nice

packages/wallet-sdk/src/core/provider/interface.ts
fan-zhang-sv: hmm prob there won't be any one using it externally, let's remove it?

cb-jake: Sounds good

packages/wallet-sdk/src/core/provider/interface.ts
```diff
@@ -42,10 +42,47 @@ export interface AppMetadata {
...
+  postOnboardingAction?: PostOnboardingAction;
+  /**
+   * @param onrampPrefillOptions
+   * @type {OnrampPrefillOptions}
+   * @description This option only functions when `postOnboardingAction` is set to `onramp`.
+   * - Prefills the onramp flow with the specified asset, chain, and suggested amount,
+   * allowing users to prefill their account.
+   * - Ensure the asset and chain are supported by the onramp provider
+   * (e.g., Coinbase Pay - CBPay).
+   */
```
spencerstock, line 75: add: defaults to an asset selector with 0 as the initial amount

Can we also add some help for devs in determining the right formats here, and determining what is supported by CBPay?

spencerstock (commented):
Looks good! 
1. How do we prevent dapps from mistakenly passing these in with eoa connections and then thinking something is broken when they do not make a difference?
2. I see some of the language is staying generic, i.e. other implementers could handle these functions differently. I think we might be better off expicitly coupling these to SCW and having the descriptions be clearer at the cost of future extensibility

packages/wallet-sdk/src/core/provider/interface.ts
cb-jake: Thanks good callout

fan-zhang-sv (dismissed):

packages/wallet-sdk/src/core/provider/interface.ts
cb-jake: updated with a link from cbpay docs

cb-jake:
@spencerstock 
> How do we prevent dapps from mistakenly passing these in with eoa connections and then thinking something is broken when they do not make a difference?

I think documentation makes the most sense. Inside the doc blocks I've added a Smart Wallet Only tag

>I see some of the language is staying generic, i.e. other implementers could handle these functions differently. I think we might be better off expicitly coupling these to SCW and having the descriptions be clearer at the cost of future extensibility

Agreed. I've updated the jsdocs

packages/wallet-sdk/src/CoinbaseWalletProvider.ts
```diff
@@ -27,10 +28,14 @@ export class CoinbaseWalletProvider extends ProviderEventEmitter implements Prov
     super();
     this.metadata = metadata;
     this.preference = preference;
-    this.communicator = new Communicator(keysUrl, metadata);
+
+    const url = typeof keysUrl === 'string' ? keysUrl : CB_KEYS_URL;
```
cb-jake, line 32: Removing keysUrl as an option resulted in a type error here. Now we're type checking for a string before passing to the communicator. Might be worth allowing unknown and validating inside the communicator

fan-zhang-sv: i see, actually maybe good to have a separate PR to remove keysUrl? i can take it on, P2/P3 maybe?

fan-zhang-sv: we can revert all keysUrl change in this PR? i'll tackle it separately

fan-zhang-sv (dismissed):
ok to merge also, i can take remove the usage of keysUrl in a follow up PR separately

fan-zhang-sv (approved):
ty! lfg

PR #1421, opened 2024-10-06
fix: project module resolution

jxom:
Upon assisting @wilsoncusack with Wagmi + importing a @wevm ESM-only library into the Coinbase Wallet SDK, I noticed that the repository wasn't set up properly for importing such packages. 

### What is introduced?

- addition of `type: "module"` property to `package.json`
- fixes `NodeNext` module resolution
- migrates from Jest to Vitest
- removes barrel files

### Why?

To break it down:
- Coinbase Wallet SDK is an ESM-only package, as evident in [`tsconfig.json#32`](https://github.com/coinbase/coinbase-wallet-sdk/blob/master/packages/wallet-sdk/tsconfig.base.json#L32)
- However, the SDK's `package.json` had a [non-existent `type: "module"`](https://github.com/coinbase/coinbase-wallet-sdk/compare/master...jxom:coinbase-wallet-sdk:jxom/modernize?expand=1#diff-0b284e7f45a5afeadcdcaf586abc089839f135b94b302f5135081e1788a43f35R14) (ESM flag) field, even though the package is being bundled as an ESM-only package. As the `type: "module"` property doesn't exist, TS "thinks" the project files are CJS modules, and as a result, we get an import error when we try to import an ESM-only module `The current file is a CommonJS module whose imports will produce 'require' calls; however, the referenced file is an ECMAScript module and cannot be imported with 'require'.`
- When adding `type: "module"` – TS now recognizes that our project is ESM – however, since `moduleResolution: "NodeNext"`, this means we need to add the `.js` extension to all imports (yes it is a throw-off, but [justification makes sense](https://github.com/microsoft/TypeScript/issues/49083#issuecomment-1435399267)).
- Upon adding the `type: "module"` property, this broke Jest. Jest's ESM support is nearly non-existent, so I migrated from Jest to Vitest, which is a trivial migration and makes the tests run a bit faster.
- Also [removed barrel files](https://x.com/_developit/status/1842225012092104732) (e.g. `src/core/message/index.ts`) which reduced bundle size by 10%.

**Before**

Bundle Size (gzipped): `33.38kB` 

```bash
$ yarn size
✔ Adding to empty webpack project
✔ Running JS in headless Chrome

  Size:         33.38 kB with all dependencies, minified and brotlied
```

**After**

Bundle Size (gzipped): `30.29kB`

```bash
$ yarn size
✔ Adding to empty webpack project
✔ Running JS in headless Chrome

  Size:         30.29 kB with all dependencies, minified and brotlied
```

### Next

I believe this repository can also benefit from the following quick wins:

1. Removing dependence on `Buffer` in-favor of [`TypedArray`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypedArray)s, so consumers don't have to install `Buffer` polyfills.
2. Removing EIP712 CJS modules in favor of [Viem's `hashTypedData`](https://github.com/wevm/viem/blob/main/src/utils/signature/hashTypedData.ts)
3. Migrating from ESLint/Prettier to [Biome](https://biomejs.dev/) would dramatically decrease lint times, as well as dependencies
4. Remove dependence on `tslib` to further decrease bundle size.


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>1</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>2</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

spencerstock:
Greetings Jake! We'll prioritize reviewing your work here. Thanks for helping out.

fan-zhang-sv:
just saw this update on extension rewrite from microsoft should we opt in to this? 
 https://github.com/microsoft/TypeScript/pull/59767

jxom:
@fan-zhang-sv – Perhaps! I don't think it is released yet though – unsure if it's worth waiting.

packages/wallet-sdk/package.json
```diff
@@ -11,59 +11,60 @@
...
     "sass": "^1.64.1",
-    "ts-jest": "^27.1.5",
-    "ts-node": "^10.9.1",
+    "size-limit": "^11.1.6",
     "tsc-alias": "^1.8.8",
     "tslib": "^2.6.0",
     "typescript": "^5.1.6"
-  }
+  },
+  "size-limit": [
```
cb-jake, line 63: Nice! Great alternative to tracking manually

packages/wallet-sdk/src/sign/walletlink/relay/ui/components/RedirectDialog/RedirectDialog.tsx
```diff
@@ -1,10 +1,10 @@
-import clsx from 'clsx';
-import { FunctionComponent, h, render } from 'preact';
```
cb-jake, line 2: I believe removing the h pragma will break the preact UI

packages/wallet-sdk/src/util/cipher.test.ts
```diff
@@ -56,7 +56,7 @@ describe('Cipher', () => {
 
       // Attempting to decrypt with a different shared secret
       await expect(decrypt(sharedSecretDerivedByPeer, encryptedMessage)).rejects.toThrow(
-        'Unsupported state or unable to authenticate data'
+        'The operation failed for an operation-specific reason'
```
cb-jake, line 59: What prompted the change in error messaging here? I don't see a related change. Looks like `The operation failed for an operation-specific reason` comes out of the webcrypto api

cb-jake:
@jxom Thanks for this PR. All in all everything looks good. 

A few callouts that we should look into:
* removing the preact pragma `h` from the hosted UI will likely break walletlink experience. I didn't see any alternative support for this
* The SDK Playground is in a broken state. Looks like changing the commonjs modules file extension to `cjs` might be the culprit. If the consuming app is required to change their module resolution to now support the wallet-sdk this would likely be considered a breaking change requiring a major version bump.

packages/wallet-sdk/src/util/cipher.test.ts
jxom: Think this would have been coming from the WebCrypto polyfill, which I removed.

packages/wallet-sdk/src/sign/walletlink/relay/ui/components/RedirectDialog/RedirectDialog.tsx
jxom: Unsure how this got removed, adding back in.

packages/wallet-sdk/src/util/cipher.test.ts
cb-jake: That makes sense.

jxom:
@cb-jake – thanks! made the changes. Although, I was unable to repro the `.cjs` runtime error. Can you pull the latest and try again?

> If the consuming app is required to change their module resolution to now support the wallet-sdk this would likely be considered a breaking change requiring a major version bump.

This shouldn't be the case as we aren't changing any of the build configuration. :)

packages/wallet-sdk/src/sign/walletlink/relay/ui/components/RedirectDialog/RedirectDialog.tsx
cb-jake: I think prettier might auto remove it. I have a task to fix up our eslint/prettier. Looks like you suggested biome. Will take a look into that.

cb-jake:
Thanks @jxom Taking a look now.

cb-jake:
@jxom changes look good. Running CI now. Walletlink/SCW interactions worked as expected.

cb-jake:
@jxom looks like we need to update CI. The `main.yml` is still referencing `compile-assets.js` vs `compile-assets.cjs`

fan-zhang-sv (commented):
Conducted manual testing via playground, smart wallet and walletlink connection both work as expected in major CUJs, including signUp/signIn, signMsg/Tx/Calls, etc

jxom:
@cb-jake pushed!

cb-jake:
Thanks @jxom! Looks like there is a merge conflict (sorry about that).

jxom:
np, just resolved

vishnumad:
@jxom This looks great, looks like the imports in `validatePreferences.ts` and `validatePreferences.test.ts` files need to be updated.

cb-jake:
@jxon thanks again for all your work on this. It looks like older version of node are missing `crypto.subtle.generateKey` inside our tests. We may need to apply a mock for older node versions.

jxom:
Do you think we should remove the 16.x / 18.x tests? Or are there Node.js consumers of `coinbase-wallet-sdk`? 

Note: Node 16 is now end-of-life, and Node 18 will be end-of-life ~Apr 2025.

cb-jake:
With the exception of SSR apps, theres probably no other nodejs consumers. We can safely remove support for Node 16. I'm on the fence with 18 since its still in active LTS. @fan-zhang-sv @spencerstock What are your thoughts on removing the test suites for Node 16 and 18?

https://nodejs.org/en/about/previous-releases
image

fan-zhang-sv:
I'm aligned on removing Node 16, but think im also leaning towards keep supporting Node 18 til end of life, to maximize the support for our consumer.

jxom:
Can take a look at adding the polyfill for the tests back in.

spencerstock:
> With the exception of SSR apps, theres probably no other nodejs consumers. We can safely remove support for Node 16. I'm on the fence with 18 since its still in active LTS. @fan-zhang-sv @spencerstock What are your thoughts on removing the test suites for Node 16 and 18? We can replace them with v22 and v24 (In a separate PR)
> 

Clean lines should be drawn here.
1. I’m not particularly concerned about dropping support for older Node versions, provided that the broader industry is also moving towards using these newer APIs.

2. If we don't fully support a version of Node we should remove the suite instead of filling with a mock or polyfill so that we aren't propping up test cases

3. Do we re-evaluate a node.js version test matrix entirely (maybe we remove it)? the SDK's functions aren't usually involved in the SSR process itself but after the page is fully loaded/hydrated.

@cb-jake what do you recommend?

cb-jake:
@spencerstock As I think about it, the more inclined I am to remove test support for 16 and 18. We can update the matrix to include 22 and 24. Internally we can add in a separate PR.

@jxom I have a PR to remove support for 16 and 18. Once we land this, you can rebase and we'll get these changes landed. https://github.com/coinbase/coinbase-wallet-sdk/pull/1424

cb-jake:
@jxom https://github.com/coinbase/coinbase-wallet-sdk/pull/1424 is merged. If you want to rebase. We should be able to get this PR landed. Thanks again!

cb-jake:
I've a [PR](https://github.com/coinbase/coinbase-wallet-sdk/pull/1428) that gets this PR into a state that can be merged. Thanks again @jxom. Appreciate the work on this! Once that lands I'll close this PR.

PR #1427, opened 2024-10-14
Web SDK sending info via searchParams

fan-zhang-sv:
### _Summary_

This PR
- updates coop check to return compatibility result
- sends over app info to SCW FE via search params for better understanding window.opener reliability issue, including
    - SDK name, version
    - app origin
    - coop compatibility

Unit Test + Manually verified work as expected

https://github.com/user-attachments/assets/1a71a599-2fdb-4d6e-9ec2-4ac3255cb4a9


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

packages/wallet-sdk/src/util/web.ts
```diff
@@ -27,8 +34,15 @@ export function closePopup(popup: Window | null) {
...
- * TODO: consolidate all UI related helper functions,
- * ones making window.xxx() document.yyy() calls.
- * e.g. WLMobileRelayUI, WalletLinkRelay, ...
- */
+async function appendAppInfoQueryParams(url: URL) {
+  const params = {
+    sdkName: LIB_NAME,
+    sdkVersion: LIB_VERSION,
+    origin: window.location.origin,
+    coopIncompatibility: await checkCrossOriginOpenerPolicyCompatibility(),
```
cb-jake, line 42: should we just pass what policy the site uses? then we can let the popup decide what to do with it. Also this naming is a bit weird. has a double negative. `coopIncompatibility === false` means its in compatible

fan-zhang-sv: sgsg, pushed an update to make `checkCrossOriginOpenerPolicy` return policy itself, and passing policy to FE

packages/wallet-sdk/src/libInfo.ts
```diff
@@ -0,0 +1,2 @@
+export const LIB_VERSION = '4.1.0';
```
cb-jake, line 1: can we rename this file to `sdk-info.ts`

packages/wallet-sdk/src/libInfo.ts
```diff
@@ -0,0 +1,2 @@
...
+export const LIB_NAME = '@coinbase/wallet-sdk';
```
cb-jake, line 2: Does removing the version.ts file break some of the versioning scripts?

packages/wallet-sdk/src/sdk-info.ts
```diff
@@ -0,0 +1,2 @@
+export const LIB_VERSION = '4.1.0';
```
cb-jake, line 1: also, can we remove the `LIB_` prefix seems redundant

packages/wallet-sdk/src/util/provider.ts
```diff
@@ -12,7 +12,11 @@ export async function fetchRPCRequest(request: RequestArguments, rpcUrl: string)
     method: 'POST',
     body: JSON.stringify(requestBody),
     mode: 'cors',
-    headers: { 'Content-Type': 'application/json', 'X-Cbw-Sdk-Version': LIB_VERSION },
+    headers: {
+      'Content-Type': 'application/json',
+      'X-Cbw-Sdk-Version': LIB_VERSION,
+      'X-Cbw-Sdk-Platform': LIB_NAME,
```
cb-jake, line 18: should this be `X-Cbw-Sdk-Name` for consistent naming?

cb-jake: Also okay renaming the LIB_NAME var

packages/wallet-sdk/src/util/web.ts
```diff
@@ -1,23 +1,30 @@
...
 const POPUP_HEIGHT = 540;
 
 // Window Management
 
-export function openPopup(url: URL): Window {
+export async function openPopup(url: URL): Promise<Window> {
   const left = (window.innerWidth - POPUP_WIDTH) / 2 + window.screenX;
   const top = (window.innerHeight - POPUP_HEIGHT) / 2 + window.screenY;
 
+  await appendAppInfoQueryParams(url);
```
cb-jake, line 14: Is there any noticeable latency added to each request? Also do these query params get appended to each popup request?

packages/wallet-sdk/src/util/checkCrossOriginOpenerPolicy.test.ts
```diff
@@ -1,17 +1,17 @@
...
+import { checkCrossOriginOpenerPolicy } from './checkCrossOriginOpenerPolicy';
 
 describe('checkCrossOriginOpenerPolicy', () => {
   beforeEach(() => {
     // Clear all mocks before each test
     jest.clearAllMocks();
   });
 
-  it('should not run if window is undefined', () => {
+  it('should return true if window is undefined', async () => {
```
vishnumad, line 9: Update test names to not have `return true` and `return false`, since the checkCrossOriginOpenerPolicy function now returns the actual policy

packages/wallet-sdk/src/sdk-info.ts
vishnumad, line 1: The `yarn prebuild` command updates this field in `version.ts` at build time. That'll need to be updated to point to `sdk-info.ts`.

packages/wallet-sdk/src/libInfo.ts
fan-zhang-sv: updated!

packages/wallet-sdk/src/libInfo.ts
fan-zhang-sv: updated script

packages/wallet-sdk/src/sdk-info.ts
fan-zhang-sv: removed LIB for both

packages/wallet-sdk/src/util/provider.ts
fan-zhang-sv: this header param is for BE to distinguish web/mobile SDKs, updating this name may need BE and mobile sdk update, let's keep it as it is?

packages/wallet-sdk/src/util/web.ts
fan-zhang-sv: good call out! this was causing safari blocking the popup, switched to saving the fetched value in closure, verified this is not blocking mobile safari

packages/wallet-sdk/src/util/checkCrossOriginOpenerPolicy.ts
```diff
@@ -0,0 +1,27 @@
...
+
+/**
+ * Checks the compatibility of the Cross-Origin-Opener-Policy (COOP) for the current window.
+ *
+ * This function fetches the current origin and examines the 'Cross-Origin-Opener-Policy' header.
+ * If the policy is set to 'same-origin', it logs an error message and returns `false`.
+ *
+ * @returns {Promise<boolean>} A promise that resolves to `true` if the COOP is compatible or if the code is running in a non-browser environment, and `false` otherwise.
+ */
+export async function checkCrossOriginOpenerPolicy(): Promise<string> {
```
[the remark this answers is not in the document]
fan-zhang-sv: updated!

packages/wallet-sdk/src/sdk-info.ts
fan-zhang-sv: updated script!

packages/wallet-sdk/src/util/web.ts
fan-zhang-sv: also, yea, params will get appended for each popup request, but the new pattern's latency would be trivial

packages/wallet-sdk/src/util/checkCrossOriginOpenerPolicy.ts
```diff
@@ -0,0 +1,49 @@
...
+ * If in a browser environment, fetches the policy from the current origin.
+ * Logs an error if the policy is 'same-origin'.
+ */
+const createCoopChecker = () => {
+  let crossOriginOpenerPolicy: string | undefined;
+
+  return {
+    getCrossOriginOpenerPolicy: () => {
+      if (crossOriginOpenerPolicy === undefined) {
+        throw new Error('Cross-Origin-Opener-Policy has not been checked yet');
```
cb-jake, line 26: should we be throwing? typical closure get pattern would be to fetch here if the values not set

fan-zhang-sv: yea this is a bit anti pattern. switched to simply return `undefined`, this is analytic code, we really shouldn't be blocking for it

fan-zhang-sv: we can't fetch either cause in my testing, ios safari will block the popup for we have too many logic between user tap and popup

packages/wallet-sdk/src/util/checkCrossOriginOpenerPolicy.ts
```diff
@@ -0,0 +1,62 @@
...
+ * If in a browser environment, fetches the policy from the current origin.
+ * Logs an error if the policy is 'same-origin'.
+ */
+const createCoopChecker = () => {
+  let crossOriginOpenerPolicy: string | undefined;
+
+  return {
+    getCrossOriginOpenerPolicy: () => {
+      if (crossOriginOpenerPolicy === undefined) {
+        return 'undefined';
```
cb-jake, line 26: why return a string value here? 

```suggestion
 if (crossOriginOpenerPolicy === undefined) {
        return 'undefined';
}
```

fan-zhang-sv: heyy correct me if im wrong, `null` is a valid value for COOP? tested in playground, we are getting null, think here i would like to distinguish null and undefined (which is more likely an error state where checkCOOP is not called/finished)

packages/wallet-sdk/src/util/provider.ts
cb-jake: sounds good

cb-jake:
@fan-zhang-sv will you add a demo video to the PR please? Thank you.

fan-zhang-sv:
@cb-jake included demo in description 🙏

cb-jake (approved):

PR #1452, opened 2024-12-03
Send app location in WalletLink request

vishnumad:
### _Summary_
This PR adds the app's `window.location` to the WalletLink event. We already do this for the origin, but there's cases where Wallet needs the full path.

Tested manually and verified that adding the field doesn't cause issues with older versions of Wallet.


fan-zhang-sv (approved):

dropkick1325 (commented):
[]()

dropkick1325:
[]()

PR #1475, opened 2025-01-09
[BW-857] Abstract handshake into `coinbase_handshake`

fan-zhang-sv:
### _Summary_

We used to see `eth_requestAccounts` as the handshake method, which does 2 things essentially:
- Authorizing the connection (by exchanging keys) and
- Returning the address

When exploring `wallet_sign`(`wallet_sendCalls`) and `wallet_connect`, the original idea was to merge another signing request with `eth_requestAccounts`, so we do 3 things at a time
- Authorizing the connection (by exchanging keys) and
- Returning the address
- Signing something

However, since the first request is for requesting trusted connection, the request itself is unencrypted. Hence, due to security reason, we couldn't request signing in this untrusted context.

This PR, is using @lukasrosario's idea, to create a `"Light Handshake"`, which will be used for establish a trusted communication channel, so subsequent requests can be transmitted in a trusted channel. In another word, instead, we are abstracting the connection authorization out from `eth_requestAccounts` to be its own request. Due to this is a concept used in Coinbase Wallet SDK only, we are naming it as `coinbase_handshake` tentatively.

#### Formula
- `eth_requestAccounts` = `coinbase_handshake` + request account effect
- ephemeral `wallet_sign` = `coinbase_handshake` + `personal_sign`
- ephemeral `wallet_sendCalls` = `coinbase_handshake` + `wallet_sendCalls`
- `wallet_connect` =  `coinbase_handshake` + wallet connect defined effect

Tested this works on desktop Chrome + mobile Safari

| Desktop Chrome | Mobile Safari |
| ---------- | --------- |
| <video src="https://github.com/user-attachments/assets/02ce9297-f126-426d-ba96-7d485c52dac3" /> | <video src="https://github.com/user-attachments/assets/d0b60db9-c908-4b04-b3dd-c6e0ffd762ba" /> |

### DIFF
--- a/packages/wallet-sdk/src/CoinbaseWalletProvider.ts
+++ b/packages/wallet-sdk/src/CoinbaseWalletProvider.ts
@@ -1,6 +1,7 @@
 import { Signer } from './sign/interface.js';
 import { createSigner, fetchSignerType, loadSignerType, storeSignerType } from './sign/util.js';
 import { Communicator } from ':core/communicator/Communicator.js';
+import { CB_WALLET_RPC_URL } from ':core/constants.js';
 import { standardErrorCodes } from ':core/error/constants.js';
 import { standardErrors } from ':core/error/errors.js';
 import { serializeError } from ':core/error/serialize.js';
@@ -15,7 +16,7 @@ import {
 } from ':core/provider/interface.js';
 import { ScopedLocalStorage } from ':core/storage/ScopedLocalStorage.js';
 import { hexStringFromNumber } from ':core/type/util.js';
-import { checkErrorForInvalidRequestArgs } from ':util/provider.js';
+import { checkErrorForInvalidRequestArgs, fetchRPCRequest } from ':util/provider.js';
 
 export class CoinbaseWalletProvider extends ProviderEventEmitter implements ProviderInterface {
   private readonly metadata: AppMetadata;
@@ -61,6 +62,8 @@ export class CoinbaseWalletProvider extends ProviderEventEmitter implements Prov
             await ephemeralSigner.cleanup(); // clean up (rotate) the ephemeral session keys
             return result as T;
           }
+          case 'wallet_getCallsStatus':
+            return fetchRPCRequest(args, CB_WALLET_RPC_URL);
           case 'net_version':
             return 1 as T; // default value
           case 'eth_chainId':
--- a/packages/wallet-sdk/src/CoinbaseWalletProvider.ts
+++ b/packages/wallet-sdk/src/CoinbaseWalletProvider.ts
@@ -53,6 +53,13 @@ export class CoinbaseWalletProvider extends ProviderEventEmitter implements Prov
             storeSignerType(signerType);
             break;
           }
+          case 'wallet_sign': {
+            const ephemeralSigner = this.initSigner('scw');
+            await ephemeralSigner.handshake({ method: 'coinbase_handshake' }); // exchange session keys
+            const result = await ephemeralSigner.request(args); // send diffie-hellman encrypted request
+            await ephemeralSigner.cleanup(); // clean up (rotate) the ephemeral session keys
+            return result as T;
+          }
           case 'net_version':
             return 1 as T; // default value
           case 'eth_chainId':
@@ -64,7 +71,7 @@ export class CoinbaseWalletProvider extends ProviderEventEmitter implements Prov
           }
         }
       }
-      return this.signer.request(args);
+      return await this.signer.request(args);
     } catch (error) {
       const { code } = error as { code?: number };
       if (code === standardErrorCodes.provider.unauthorized) this.disconnect();
--- a/packages/wallet-sdk/src/CoinbaseWalletProvider.ts
+++ b/packages/wallet-sdk/src/CoinbaseWalletProvider.ts
@@ -53,6 +53,7 @@ export class CoinbaseWalletProvider extends ProviderEventEmitter implements Prov
             storeSignerType(signerType);
             break;
           }
+          case 'wallet_sendCalls':
           case 'wallet_sign': {
             const ephemeralSigner = this.initSigner('scw');
             await ephemeralSigner.handshake({ method: 'coinbase_handshake' }); // exchange session keys
--- a/packages/wallet-sdk/src/sign/scw/SCWSigner.ts
+++ b/packages/wallet-sdk/src/sign/scw/SCWSigner.ts
@@ -124,6 +124,7 @@ export class SCWSigner implements Signer {
         return this.handleSwitchChainRequest(request);
       case 'eth_ecRecover':
       case 'personal_sign':
+      case 'wallet_sign':
       case 'personal_ecRecover':
       case 'eth_signTransaction':
       case 'eth_sendTransaction':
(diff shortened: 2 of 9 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

packages/wallet-sdk/src/CoinbaseWalletProvider.ts
```diff
@@ -53,6 +53,14 @@ export class CoinbaseWalletProvider extends ProviderEventEmitter implements Prov
...
             break;
           }
+          case 'wallet_sendCalls':
+          case 'wallet_sign': {
+            const ephemeralSigner = this.initSigner('scw');
+            await ephemeralSigner.handshake({ method: 'coinbase_handshake' }); // exchange session keys
+            const result = await ephemeralSigner.request(args); // send diffie-hellman encrypted request
+            await ephemeralSigner.cleanup(); // clean up (rotate) the ephemeral session keys
+            return result as T;
+          }
```
fan-zhang-sv, line 63: we are going to do similar thing for `wallet_connect`, just that the signer won't be ephemeral but will have some effect in SCWSigner.handshake

packages/wallet-sdk/src/CoinbaseWalletProvider.ts
```diff
@@ -64,7 +72,7 @@ export class CoinbaseWalletProvider extends ProviderEventEmitter implements Prov
           }
         }
       }
-      return this.signer.request(args);
+      return await this.signer.request(args);
```
fan-zhang-sv, line 75: await here so we can catch the unauthorized error below

packages/wallet-sdk/src/CoinbaseWalletProvider.ts
```diff
@@ -53,6 +54,16 @@ export class CoinbaseWalletProvider extends ProviderEventEmitter implements Prov
             storeSignerType(signerType);
             break;
           }
+          case 'wallet_sendCalls':
```
cb-jake, line 57: What happens if an unconnected EOA calls these methods. i.e. I have the extension

packages/wallet-sdk/src/sign/scw/SCWSigner.ts
```diff
@@ -61,6 +61,10 @@ export class SCWSigner implements Signer {
   }
 
   async handshake(args: RequestArguments) {
+    // Open the popup before constructing the request message.
+    // This is to ensure that the popup is not blocked by some browsers (i.e. Safari)
+    await this.communicator.waitForPopupLoaded?.();
+
```
cb-jake, line 67: how does this impact the UX. Do we see longer loading times on the popup for the key exchange?

fan-zhang-sv: theoretically yea, but in my local testing, i couldn't notice the difference. this pattern is already being used for regular request, where it requires message encryption and also early loaded the popup to bypass popup blocking, at [here](https://github.com/coinbase/coinbase-wallet-sdk/blob/felix/coinbase_handshake/packages/wallet-sdk/src/sign/scw/SCWSigner.ts#L147C1-L150C52)

packages/wallet-sdk/src/CoinbaseWalletProvider.ts
fan-zhang-sv: tested locally, these rpc methods are currently not supported for EOA and in extension. the request will be passed over to extension, but extension will reject with unsupported error at serviceWorker lever
image

cb-jake: Great!

packages/wallet-sdk/src/CoinbaseWalletProvider.ts
```diff
@@ -53,6 +54,16 @@ export class CoinbaseWalletProvider extends ProviderEventEmitter implements Prov
...
+          case 'wallet_sign': {
+            const ephemeralSigner = this.initSigner('scw');
+            await ephemeralSigner.handshake({ method: 'coinbase_handshake' }); // exchange session keys
+            const result = await ephemeralSigner.request(args); // send diffie-hellman encrypted request
+            await ephemeralSigner.cleanup(); // clean up (rotate) the ephemeral session keys
+            return result as T;
+          }
```
cb-jake, line 64: I wonder if we should move this into a separate function/method to keep the request function small and make it easier to test?

cb-jake: Also we should add tests for this method

packages/wallet-sdk/src/sign/scw/SCWSigner.ts
cb-jake: Does this impact `eth_requestAccounts` at all?

fan-zhang-sv: theoretically yes, for `eth_requestAccounts`, by doing this, we are loading popup a bit sooner. 

the time delta would be considered very small

Before
1. User clicking the button
2. SDK prepping handshake message envelope
3. SDK opens popup
4. SDK pass envelope to popup

The time delta between 1 and 3 is the time used for prepping message envelope, this time is considered very small, that it does not trigger mobile browser popup blocking policy.

After
1. User clicking the button
2. SDK opens popup
3. SDK prepping handshake message envelope
4. SDK pass envelope to popup

so I think moving 3 to be between popup opening and passing envelope to popup won't have a big impact on UX

packages/wallet-sdk/src/CoinbaseWalletProvider.ts
fan-zhang-sv: sg added tests on ephemeral signer behaviour!

hmmm i'm actually kinda leaning towards keeping the ephemeral signer logic in place, cause this is the only place that we will ever call this logic, so i'm feeling a bit hesitant on abstracting it to a separate func, wdyt

cb-jake: Sounds good for now.

cb-jake (approved):

PR #1490, opened 2025-01-28
Add support for sub account and the new addAddress RPC

cb-jake:
### _Summary_

This PR is a bit larger (sorry)

Includes: 

* creating a sub account
* addAddress RPC support
*


packages/wallet-sdk/src/features/clients/constants.ts
```diff
@@ -0,0 +1,13 @@
...
+/**
+ * TODO[jake]: ensure all supported chains are added here
+ */
+
+// Currently only testnets are supported
+export const SUPPORTED_CHAIN_MAP = {
+  [baseSepolia.id]: baseSepolia,
+  [sepolia.id]: sepolia,
+  [optimismSepolia.id]: optimismSepolia,
+} as Record<number, Chain>;
```
fan-zhang-sv, line 13: we prob don't want to introduce any blocking logic into SDK, just in case in the future we need to add support to certain chain, we don't block older SDK

i wonder if this supported maps and the if supported check is needed, given the SCWChain[] is returned by SCW FE and should always be valid

packages/wallet-sdk/src/features/sub-accounts/createSubAccountSigner.ts
```diff
@@ -0,0 +1,109 @@
...
+  }
+
+  const signerAddress = await getAddress();
+  const account = toViemAccount({
+    type,
+    sign,
+    address: signerAddress,
+  });
+
+  const client = getClient(subAccount.chainId ?? baseSepolia.id);
```
fan-zhang-sv, line 37: can we add a comment to remind us to change this to base when we release this feature

packages/wallet-sdk/src/sign/scw/SCWSigner.ts
```diff
@@ -244,29 +269,106 @@ export class SCWSigner implements Signer {
...
+    this.subAccount = getSubAccountFromStorage();
+    return response;
+  }
+
+  private shouldRequestUseSubAccountSigner(request: RequestArguments) {
+    const sender = getSenderFromRequest(request);
+    // if the sender is undefined, it means the application did not provide a sender
+    // in this case, we assume the request is for the active sub account (if present)
+    // if the sender is defined, we check if it is the same as the active sub account
+    // if not, we use the root account signer
```
fan-zhang-sv, line 353: hmm i wonder if we should reject with standardErrors.invalidParam instead

packages/wallet-sdk/src/sign/scw/SCWSigner.ts
```diff
@@ -244,29 +269,106 @@ export class SCWSigner implements Signer {
...
+    return (
+      (this.subAccount && sender === undefined) ||
+      (this.subAccount && this.subAccount.address === sender)
+    );
+  }
+
+  private async sendRequestSubAccountSigner(request: RequestArguments) {
```
fan-zhang-sv, line 360: nit: i personally think this name is a bit confusing, wonder if we should do `signRequestWithSubAccountSigner`, or simply `sendRequestToSubAccountSigner`

packages/wallet-sdk/src/features/clients/constants.ts
cb-jake: Yeah I was torn on this. Its a great point. We need to be able to load the clients based on the chains and viem has some strict types. We might be able to create the clients based on just the rpc url. We do want to limit to testnet only for our first release

packages/wallet-sdk/src/features/sub-accounts/createSubAccountSigner.ts
cb-jake: good callout

packages/wallet-sdk/src/sign/scw/SCWSigner.ts
cb-jake: Yes `sendRequestToSubAccountSigner` is better. Will update

examples/testapp/src/pages/sub-accounts/components/AddAddressButton.tsx
```diff
@@ -0,0 +1,47 @@
...
+  const handleAddAddress = useCallback(async () => {
+    if (!sdk) {
+      return;
+    }
+    const provider = sdk.getProvider();
+    const response = (await provider.request({
+      method: 'wallet_addAddress',
+      params: [
+        {
+          chainId: 84532,
```
wilsoncusack, line 17: Idea being this address is limited to one chain, correct?

packages/wallet-sdk/src/features/sub-accounts/types.ts
```diff
@@ -0,0 +1,34 @@
...
+import type { SignReturnType as WebAuthnSignReturnType } from 'webauthn-p256';
+
+export type SubAccountSigner = {
+  type: 'webAuthn'; // TODO: add local support
+  /**
+   * Signs a raw payload
+   * @param rawPayload - The raw payload to sign
+   * @returns {Hex} Signed payload
+   */
+  getSigner: () => Promise<
```
wilsoncusack, line 12: can we consolidate the types here? Just like a Viem account? Or what is webauthn different than normal

wilsoncusack: @jxom any take? :D

examples/testapp/src/pages/sub-accounts/components/AddAddressButton.tsx
cb-jake: Yes the idea was each account is per chain and we wont sync across chains

packages/wallet-sdk/src/features/sub-accounts/types.ts
cb-jake: I've updated this to resolve to a viem account, this removes the need to getAddress and type methods

new interface 

```tsx
getSigner: () => Promise<LocalAccount|WebAuthnAccount>
```

cb-jake: > can we consolidate the types here? Just like a Viem account? Or what is webauthn different than normal

The return signature types are different. If we want to keep this generic we'll likely need our sign function to take a Hex and resolve to a Hex and then internally map to the various account types i.e. Local or WebAuthn

PR #1554, opened 2025-03-13
feat: prepare wallet rpcs

stephancill:
### _Summary_

Replace wallet_sendCalls with wallet_prepareCalls + wallet_sendPreparedCalls

Tried Send Calls on playground and updated tests

### DIFF
--- a/examples/testapp/src/pages/add-sub-account/index.page.tsx
+++ b/examples/testapp/src/pages/add-sub-account/index.page.tsx
@@ -9,8 +9,7 @@ import {
 } from '@chakra-ui/react';
 import { createCoinbaseWalletSDK, getCryptoKeyAccount } from '@coinbase/wallet-sdk';
 import { useEffect, useState } from 'react';
-
-import { mnemonicToAccount } from 'viem/accounts';
+import { generatePrivateKey, privateKeyToAccount } from 'viem/accounts';
 import { AddOwner } from './components/AddOwner';
 import { AddSubAccount } from './components/AddSubAccount';
 import { Connect } from './components/Connect';
@@ -45,11 +44,16 @@ export default function SubAccounts() {
     const getSigner =
       signerType === 'cryptokey'
         ? getCryptoKeyAccount
-        : async () => ({
-            account: mnemonicToAccount(
-              'test test test test test test test test test test test junk'
-            ),
-          });
+        : async () => {
+            let privateKey = localStorage.getItem('secp256k1-private-key') as `0x${string}` | null;
+            if (!privateKey) {
+              privateKey = generatePrivateKey();
+              localStorage.setItem('secp256k1-private-key', privateKey);
+            }
+            return {
+              account: privateKeyToAccount(privateKey),
+            };
+          };
 
     setGetSubAccountSigner(() => getSigner);
   }, [signerType]);
--- a/examples/testapp/src/pages/add-sub-account/index.page.tsx
+++ b/examples/testapp/src/pages/add-sub-account/index.page.tsx
@@ -1,7 +1,16 @@
-import { Container, VStack } from '@chakra-ui/react';
+import {
+  Container,
+  FormControl,
+  FormLabel,
+  Radio,
+  RadioGroup,
+  Stack,
+  VStack,
+} from '@chakra-ui/react';
 import { createCoinbaseWalletSDK, getCryptoKeyAccount } from '@coinbase/wallet-sdk';
 import { useEffect, useState } from 'react';
 
+import { mnemonicToAccount } from 'viem/accounts';
 import { AddOwner } from './components/AddOwner';
 import { AddSubAccount } from './components/AddSubAccount';
 import { Connect } from './components/Connect';
@@ -11,9 +20,39 @@ import { PersonalSign } from './components/PersonalSign';
 import { SendCalls } from './components/SendCalls';
 import { SpendPermissions } from './components/SpendPermissions';
 
+type SignerType = 'cryptokey' | 'secp256k1';
+
 export default function SubAccounts() {
   const [sdk, setSDK] = useState<ReturnType<typeof createCoinbaseWalletSDK>>();
   const [subAccountAddress, setSubAccountAddress] = useState<string>();
+  const [signerType, setSignerType] = useState<SignerType>('cryptokey');
+  const [getSubAccountSigner, setGetSubAccountSigner] = useState<typeof getCryptoKeyAccount>(
+    () => getCryptoKeyAccount
+  );
+
+  useEffect(() => {
+    const stored = localStorage.getItem('signer-type');
+    if (stored !== null) {
+      setSignerType(stored as SignerType);
+    }
+  }, []);
+
+  useEffect(() => {
+    localStorage.setItem('signer-type', signerType);
+  }, [signerType]);
+
+  useEffect(() => {
+    const getSigner =
+      signerType === 'cryptokey'
+        ? getCryptoKeyAccount
+        : async () => ({
+            account: mnemonicToAccount(
+              'test test test test test test test test test test test junk'
+            ),
+          });
+
+    setGetSubAccountSigner(() => getSigner);
+  }, [signerType]);
 
   useEffect(() => {
     const sdk = createCoinbaseWalletSDK({
@@ -22,12 +61,8 @@ export default function SubAccounts() {
         keysUrl: 'http://localhost:3005/connect',
         options: 'smartWalletOnly',
       },
-      toSubAccountSigner: getCryptoKeyAccount,
+      toSubAccountSigner: getSubAccountSigner,
     });
-
-    if (!sdk) {
-      return;
-    }
 
     setSDK(sdk);
     const provider = sdk.getProvider();
@@ -35,13 +70,26 @@ export default function SubAccounts() {
     provider.on('accountsChanged', (accounts) => {
       console.info('accountsChanged', accounts);
     });
-  }, []);
+  }, [getSubAccountSigner]);
 
   return (
     <Container mb={16}>
       <VStack w="full" spacing={4}>
+        <FormControl>
+          <FormLabel>Select Signer Type</FormLabel>
+          <RadioGroup value={signerType} onChange={(value: SignerType) => setSignerType(value)}>
+            <Stack direction="row">
+              <Radio value="cryptokey">CryptoKey</Radio>
+              <Radio value="secp256k1">secp256k1</Radio>
+            </Stack>
+          </RadioGroup>
+        </FormControl>
         <Connect sdk={sdk} />
-        <AddSubAccount sdk={sdk} onAddSubAccount={setSubAccountAddress} />
+        <AddSubAccount
+          sdk={sdk}
+          onAddSubAccount={setSubAccountAddress}
+          signerFn={getSubAccountSigner}
+        />
         <PersonalSign sdk={sdk} subAccountAddress={subAccountAddress} />
         <SendCalls sdk={sdk} subAccountAddress={subAccountAddress} />
         <GrantSpendPermission sdk={sdk} subAccountAddress={subAccountAddress} />
--- a/examples/testapp/src/pages/add-sub-account/index.page.tsx
+++ b/examples/testapp/src/pages/add-sub-account/index.page.tsx
@@ -45,10 +45,10 @@ export default function SubAccounts() {
       signerType === 'cryptokey'
         ? getCryptoKeyAccount
         : async () => {
-            let privateKey = localStorage.getItem('secp256k1-private-key') as `0x${string}` | null;
+            let privateKey = localStorage.getItem('cbwsdk.demo.add-sub-account.pk') as `0x${string}` | null;
             if (!privateKey) {
               privateKey = generatePrivateKey();
-              localStorage.setItem('secp256k1-private-key', privateKey);
+              localStorage.setItem('cbwsdk.demo.add-sub-account.pk', privateKey);
             }
             return {
               account: privateKeyToAccount(privateKey),
--- a/packages/wallet-sdk/src/core/rpc/wallet_sendPreparedCalls.ts
+++ b/packages/wallet-sdk/src/core/rpc/wallet_sendPreparedCalls.ts
@@ -43,6 +43,7 @@ export type SendPreparedCallsParams = [
 export type SendPreparedCallsReturnValue = string[];
 
 export type SendPreparedCallsSchema = {
+  Method: 'wallet_sendPreparedCalls';
   Parameters: SendPreparedCallsParams;
   ReturnType: SendPreparedCallsReturnValue;
 };
--- a/packages/wallet-sdk/src/core/rpc/wallet_sendPreparedCalls.ts
+++ b/packages/wallet-sdk/src/core/rpc/wallet_sendPreparedCalls.ts
@@ -1,3 +1,5 @@
+import { Hex } from 'viem';
+
 export type WebauthnSignatureType = {
   type: 'webauthn';
   data: {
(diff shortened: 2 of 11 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

packages/wallet-sdk/src/sign/scw/utils/createSubAccountSigner.ts
```diff
@@ -52,74 +70,198 @@ export async function createSubAccountSigner({ chainId }: { chainId: number }) {
...
+          throw standardErrors.rpc.invalidParams('chainId is required');
         }
-        case 'personal_sign': {
-          assertArrayPresence(args.params);
-          return account.signMessage({ message: args.params[0] } as {
-            message: SignableMessage;
-          });
+
+        const walletClient = getClient(hexToNumber(chainIdHex));
+        assertPresence(walletClient, standardErrors.rpc.internal('wallet client not found'));
```
montycheese, line 227: for chain {chainIdRaw}

packages/wallet-sdk/src/sign/scw/utils/createSubAccountSigner.ts
```diff
@@ -52,74 +70,198 @@ export async function createSubAccountSigner({ chainId }: { chainId: number }) {
...
+        assertPresence(chainId, standardErrors.rpc.invalidParams('chainId is required'));
+
+        const walletClient = getClient(chainId);
+        assertPresence(walletClient, standardErrors.rpc.internal('wallet client not found'));
+
+        if (!args.params[0]) {
+          throw standardErrors.rpc.invalidParams('params are required');
+        }
+
+        if (!('calls' in args.params[0])) {
```
montycheese, line 104: seems like the current code pattern is to use `get()` for deep access. Let's keep for consistency
```suggestion
        if (!get(args.params[0], 'calls')) {
```

packages/wallet-sdk/src/sign/scw/utils/createSubAccountSigner.ts
```diff
@@ -52,74 +70,198 @@ export async function createSubAccountSigner({ chainId }: { chainId: number }) {
...
+
+        if (owner!.address && isAddress(owner!.address)) {
+          signatureData = {
+            type: 'secp256k1',
+            data: {
+              address: owner!.address,
+              signature: signature as Hex,
+            },
+          };
+        } else if (typeof signature === 'object' && 'webauthn' in signature) {
```
montycheese, line 146:
```suggestion
        } else if (!get(signature, 'webauthn')) {
```

packages/wallet-sdk/src/sign/scw/utils/createSubAccountSigner.ts
```diff
@@ -52,74 +70,198 @@ export async function createSubAccountSigner({ chainId }: { chainId: number }) {
...
+        assertPresence(walletClient, standardErrors.rpc.internal('wallet client not found'));
+
+        if (!args.params[0]) {
+          throw standardErrors.rpc.invalidParams('params are required');
         }
-        case 'eth_signTypedData_v4': {
-          assertArrayPresence(args.params);
-          return account.signTypedData(args.params[1] as TypedDataDefinition);
+
+        if (!('calls' in args.params[0])) {
```
montycheese, line 233: nit: use `get`

montycheese (commented):
Looks good so far, a few comments and questions

examples/testapp/src/pages/add-sub-account/index.page.tsx
```diff
@@ -11,9 +20,39 @@ import { PersonalSign } from './components/PersonalSign';
...
+  }, [signerType]);
+
+  useEffect(() => {
+    const getSigner =
+      signerType === 'cryptokey'
+        ? getCryptoKeyAccount
+        : async () => ({
+            account: mnemonicToAccount(
+              'test test test test test test test test test test test junk'
+            ),
```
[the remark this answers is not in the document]
stephancill: it will derive the same address for the signer but sub account address will be unique because of the universal address in the initcode. was just a shortcut in the interest of time, but can add storing/fetching from localstorage

packages/wallet-sdk/src/sign/scw/SCWSigner.ts
```diff
@@ -386,6 +386,8 @@ export class SCWSigner implements Signer {
     const signer = await createSubAccountSigner({
       chainId: this.chain.id,
     });
-    return signer.request(request);
+
+    // `RequestArguments` is not correctly typed for viem EIP1193RequestFn
+    return signer.request(request as any);
```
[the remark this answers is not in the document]
stephancill: it's just an internal type that is not strict enough, will fix

packages/wallet-sdk/src/sign/scw/utils/createSubAccountSigner.ts
stephancill: hm seems like `get` doesn't actually narrow the type

stephancill: managed to avoid it entirely

montycheese (dismissed):

examples/testapp/src/pages/add-sub-account/index.page.tsx
```diff
@@ -11,9 +19,44 @@ import { PersonalSign } from './components/PersonalSign';
...
+  useEffect(() => {
+    localStorage.setItem('signer-type', signerType);
+  }, [signerType]);
+
+  useEffect(() => {
+    const getSigner =
+      signerType === 'cryptokey'
+        ? getCryptoKeyAccount
+        : async () => {
+            let privateKey = localStorage.getItem('secp256k1-private-key') as `0x${string}` | null;
```
cb-jake, line 48: nit: we should prefix this with `cbwsdk.demo.<key>` so its clear that the value is from the demo app in the sdk

cb-jake: I also created a pk example for the import flow. Maybe we could reuse the pk across demos https://github.com/coinbase/coinbase-wallet-sdk/blob/master/examples/testapp/src/pages/import-sub-account/index.page.tsx#L38C35-L38C66

cb-jake: This can be a follow up PR (either you or I can take)

packages/wallet-sdk/src/core/rpc/wallet_sendPreparedCalls.ts
```diff
@@ -0,0 +1,51 @@
...
+    type: string;
+    data: any;
+    chainId: Hex;
+    signature: WebauthnSignatureType | Secp256k1SignatureType;
+  },
+];
+
+export type SendPreparedCallsReturnValue = string[];
+
+export type SendPreparedCallsSchema = {
```
cb-jake, line 47: I like having the RPC schema. In a follow up we should update `wallet_addSubAccount` and `wallet_connect`. I'll make a task to track it in linear.

packages/wallet-sdk/src/sign/scw/SCWSigner.ts
```diff
@@ -12,7 +8,7 @@ import { AppMetadata, ProviderEventCallback, RequestArguments } from ':core/prov
 import { WalletConnectResponse } from ':core/rpc/wallet_connect.js';
 import { Address } from ':core/type/index.js';
 import { ensureIntNumber, hexStringFromNumber } from ':core/type/util.js';
-import { createClients, SDKChain } from ':store/chain-clients/utils.js';
+import { SDKChain, createClients } from ':store/chain-clients/utils.js';
```
cb-jake, line 11: Interesting that this import updated. I believe we have import sorting enabled for biome

packages/wallet-sdk/src/sign/scw/utils/createSubAccountSigner.ts
```diff
@@ -52,74 +68,205 @@ export async function createSubAccountSigner({ chainId }: { chainId: number }) {
...
+      case 'wallet_addEthereumChain':
+      case 'wallet_switchEthereumChain':
+      default:
+        throw standardErrors.rpc.methodNotSupported();
+    }
+  };
+
+  return {
+    request,
+    address: subAccount.address,
```
cb-jake, line 270: Is address being used anywhere?

packages/wallet-sdk/src/util/encoding.ts
```diff
@@ -0,0 +1,55 @@
+import { hexToBytes, numberToHex, trim } from "viem";
+
+export function base64ToBase64Url(base64: string): string {
+  return base64.replaceAll('+', '-').replaceAll('/', '_').replace(/=+$/, '');
+}
+
+export function arrayBufferToBase64Url(buffer: ArrayBuffer): string {
+  // First convert to regular base64
+  const base64 = Buffer.from(buffer).toString('base64');
```
cb-jake, line 9: I don't think we should be using Buffer directly. The SDK needs to run in web context without the need to polyfill.

cb-jake:
```tsx
function arrayBufferToBase64(buffer) {
  let binary = '';
  const bytes = new Uint8Array(buffer);
  const len = bytes.byteLength;
  for (let i = 0; i < len; i++) {
    binary += String.fromCharCode(bytes[i]);
  }
  return btoa(binary);
}
```

https://stackoverflow.com/questions/9267899/how-can-i-convert-an-arraybuffer-to-a-base64-encoded-string

cb-jake (commented):
Looks good. Only thing that has to change is removing Buffer. The only reason this is working locally is the test app is next which polyfills Buffer into the browser env

packages/wallet-sdk/src/util/encoding.ts
stephancill: gotcha, should probably be a linter warning then. will look into that

examples/testapp/src/pages/add-sub-account/index.page.tsx
stephancill: Happy to take that

packages/wallet-sdk/src/sign/scw/SCWSigner.ts
stephancill: Maybe this?
```
>>> ord("S")
83
>>> ord("c")
99
```

cb-jake: Most likely. all good

packages/wallet-sdk/src/util/encoding.ts
cb-jake: Downside of biome. Still some lacking tooling. hmm...

packages/wallet-sdk/src/sign/scw/utils/createSubAccountSigner.ts
```diff
@@ -52,74 +68,204 @@ export async function createSubAccountSigner({ chainId }: { chainId: number }) {
...
+              version: '1.0',
+              type: prepareCallsResponse.type,
+              data: prepareCallsResponse.userOp,
+              chainId: prepareCallsResponse.chainId,
+              signature: signatureData,
+            },
+          ],
+        })) as SendPreparedCallsSchema['ReturnType'];
+
+        return sendPreparedCallsResponse[0];
```
cb-jake, line 189: This response doesn't map to anything. I think it might be the signature vs the user op hash

cb-jake: weird the userop hash for my test was `0x1c91c7eae34080db52cdcb7abdaf1e5f05d27247d0ce99444692c203b19bcd57` but the response returned `0x1c91c7eae34080db52cdcb7abdaf1e5f05d27247d0ce99444692c203b19bcd570000000000000000000000000000000000000000000000000000000000014a34`

packages/wallet-sdk/src/sign/scw/utils/createSubAccountSigner.ts
```diff
@@ -52,74 +68,204 @@ export async function createSubAccountSigner({ chainId }: { chainId: number }) {
...
+        const chainId = get(args.params[0], 'chainId');
+        if (!chainId) {
+          throw standardErrors.rpc.invalidParams('chainId is required');
+        }
+
+        if (!isHex(chainId)) {
+          throw standardErrors.rpc.invalidParams('chainId must be a hex encoded integer');
+        }
+
+        const walletClient = getClient(hexToNumber(chainId));
```
cb-jake, line 228: nit: these are technically public clients

packages/wallet-sdk/src/util/encoding.ts
```diff
@@ -0,0 +1,56 @@
...
+
+export function arrayBufferToBase64Url(buffer: ArrayBuffer): string {
+  // First convert to regular base64
+  const base64String = btoa(String.fromCharCode(...new Uint8Array(buffer)));
+
+  // Then convert to base64url
+  return base64ToBase64Url(base64String);
+}
+
+export function convertCredentialToJSON(credential: any) {
```
cb-jake, line 15: We should avoid type any. Either have a credential type or set it to unknown and assert the params

```tsx
export function convertCredentialToJSON(credential: unknow) {
  assertCredential(credential);
  // ... rest
}

type Credential = {
    id: string;
    rawId: string;
    response: {
      authenticatorData: AuthenticatorDat;
      clientDataJSON: ClientDataJSON;
      signature: Hex;
    },
    type: "" | "", // possible types
}

function assertCredential(credential: unknown): asserts is Credential {
  // validate and throw
}
```

stephancill: good callout, thanks. i moved the remaining preparation logic from the rpc method handler into here as well so now it takes the output from ox's `sign` and serializes it to JSON

packages/wallet-sdk/src/sign/scw/utils/createSubAccountSigner.ts
stephancill: afaik that is an arbitrary id that needs to be awaited using wallet_getCallsStatus which will return receipts

packages/wallet-sdk/src/util/encoding.ts
```diff
@@ -0,0 +1,66 @@
...
+
+export function convertCredentialToJSON({
+  webauthn,
+  signature,
+  id,
+}: {
+  signature: Hex;
+  webauthn: WebAuthnP256.SignMetadata;
+  id: string;
+}) {
```
cb-jake, line 24: Thank you!

packages/wallet-sdk/src/sign/scw/utils/createSubAccountSigner.ts
cb-jake: Yeah it should be returning the correct callId. I'd like to index on this a bit more to make sure that we're always returning the properly formatted call id. Lets talk about it in stand up

montycheese (approved):

PR #1592, opened 2025-04-04
Montana/experimental hsa

montycheese:

### DIFF
--- a/packages/wallet-sdk/src/CoinbaseWalletProvider.ts
+++ b/packages/wallet-sdk/src/CoinbaseWalletProvider.ts
@@ -18,6 +18,7 @@ import { ScopedLocalStorage } from ':core/storage/ScopedLocalStorage.js';
 import { hexStringFromNumber } from ':core/type/util.js';
 import { store } from ':store/store.js';
 import { checkErrorForInvalidRequestArgs, fetchRPCRequest } from ':util/provider.js';
+import { getCryptoKeyAccount } from './kms/crypto-key/index.js';
 import { Signer } from './sign/interface.js';
 import { createSigner, loadSignerType, storeSignerType } from './sign/util.js';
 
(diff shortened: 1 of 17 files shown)


packages/wallet-sdk/src/CoinbaseWalletProvider.ts
```diff
@@ -35,10 +52,17 @@ export class CoinbaseWalletProvider extends ProviderEventEmitter implements Prov
       preference,
     });
 
+    if (this.preference.enableAutomaticSubAccountCreation) {
```
cb-jake, line 55: @montycheese which 3rd party lib was this aiming to fix? I think we can move this up into the CoinbaseWalletSDK class vs the provider but want to make sure I test that edge case

packages/wallet-sdk/src/CoinbaseWalletProvider.ts
```diff
@@ -47,21 +71,89 @@ export class CoinbaseWalletProvider extends ProviderEventEmitter implements Prov
       if (!this.signer) {
         switch (args.method) {
           case 'eth_requestAccounts': {
-            const signerType = await this.requestSignerSelection(args);
+            // this causes a popup which we dont want.
+            //const signerType = await this.requestSignerSelection(args);
+            const signerType = 'scw';
```
cb-jake, line 76: we still need the popup so users can select whether or not they want to use a scw or walletlink

packages/wallet-sdk/src/CoinbaseWalletProvider.ts
montycheese: it was wagmi. 

it was just a way to set the default toSubAcccountSigner impl if one wasn't being provided, since wagmi doesn't currently see all of our sdk params, just the preferences map.

packages/wallet-sdk/src/CoinbaseWalletProvider.ts
montycheese: I may have just removed this in the moment because it was annoying

Issue #1860, opened 2026-01-14
Feature Request: Support for connecting multiple wallets at once


inrush-irony:
### Is your feature request related to a problem? Please describe.

Right now, the Coinbase Wallet SDK only lets you connect to one wallet at a time. It would be awesome if there was a way to connect to multiple wallets at the same time. That way, users wouldn’t have to disconnect and reconnect every time they want to switch between wallets.

### Describe the solution you'd like

It’d be great if the SDK added a feature to manage multiple wallet connections. Developers could switch between wallets in the same session without hassle.

### Why it would help
- It would make it way easier to manage multiple wallets in an app.
- Users wouldn’t have to constantly disconnect and reconnect.
- It’d make apps feel smoother when interacting with multiple wallets.

ZayanKhan-12:
Outside contributor here — I dug into why this isn't possible today, since the answer turned out to be more specific than "the SDK doesn't support it". Posting it so the request is an actionable engineering problem rather than a wish. No action needed from me.

## Why a second wallet can't connect

It isn't a missing API — instantiating the SDK twice doesn't help. Connection state lives in a single module-level store:

```ts
// packages/wallet-sdk/src/store/store.ts
export const sdkstore = createStore(persist<StoreState>(...));
```

Fifteen non-test modules import it directly: `CoinbaseWalletProvider`, `SCWSigner`, `SCWKeyManager`, `createSubAccountSigner`, `handleAddSubAccountOwner`, `assertSubAccount`, the telemetry event modules, and others. Every one reads or writes the singleton rather than receiving state.

So `createCoinbaseWalletSDK` overwrites rather than adds. Run against the current SDK:

```
appName seen by BOTH instances after creating B: Wallet B
appChainIds: [8453]
accounts after B connects: ["0xBBBB…"]
```

Two distinct SDK objects, one shared config, one account slot. The second call wins — silently.

## What it would take

Roughly, in the order it could land:

1. **Make the store creatable.** Extract the body of `store.ts` into `createSdkStore()` returning the store plus its action helpers. Keep `export const store = createSdkStore()` as the default instance so nothing breaks on day one.
2. **Give the persist middleware a per-instance key.** It's hardcoded to `name: 'cbwsdk.store'`; two instances sharing that key would fight over `localStorage`. It probably wants to derive from something stable per connection rather than per app.
[...]

PR #1912, opened 2026-09-15
feat: warn when a second SDK instance replaces the first

ZayanKhan-12:
### What this is

A scoped change toward #1860, plus the analysis behind it. It does **not** implement multi-wallet support — see the reasoning below.

### Why multiple wallets don't work today

The blocker isn't a missing API. Connection state lives in a single module-level zustand store:

```ts
// packages/wallet-sdk/src/store/store.ts
export const sdkstore = createStore(persist<StoreState>(...));
```

Fifteen non-test modules import it directly, including `SCWSigner`, `SCWKeyManager`, `createSubAccountSigner` and telemetry. So creating a second SDK doesn't give you a second connection — it replaces the first. Verified against the real SDK:

```
appName seen by BOTH instances after creating B: Wallet B
appChainIds: [8453]
accounts after B connects: ["0xBBBB…"]
```

`createCoinbaseWalletSDK` writes `store.config.set(options)` on every call, and there is one account slot.

### What this PR changes

That replacement happens **silently**, which is the worst part for anyone hitting it: two instances appear to exist, the second quietly wins, and nothing explains why the first wallet stopped working. This makes it explicit while the underlying limitation stands.

`createCoinbaseWalletSDK` now warns when called again with *different* metadata, naming both apps and linking the issue. Repeat calls with *identical* metadata stay silent — React strict mode and hot reloading both do that legitimately, and warning there would be noise.

One subtlety worth calling out: the check compares against metadata recorded for the current page session, **not** against the store. `config` is persisted to `localStorage` and rehydrated synchronously on load, so comparing against stored state would warn whenever a dapp changed its own `appName` between deploys — a false positive on first load after any release.

The README now documents that the SDK holds one connection at a time.

### What I deliberately did not do

Making this actually work means moving the store from a module singleton to per-instance state and threading it through all fifteen modules — including the signer and key manager. That's a core architectural change to wallet session and key handling, it's yours to design, and I didn't think it belonged in an unsolicited PR. I've written up the concrete shape of that refactor on #1860 so it's an actionable engineering problem rather than a wish.

Happy to implement whichever direction you pick, or to drop this entirely if you'd rather not carry a warning.

### Testing

Full CI gate, locally, from a clean checkout:

| | baseline (`master`) | this branch |
| :--- | :--- | :--- |
| `yarn lint` (biome) | 2 warnings | same 2 — no new findings |
| `yarn build:packages` | clean | clean |
| `yarn typecheck` | clean | clean |
| `yarn test` | 47 files, 454 passed, 1 skipped | 47 files, **458 passed**, 1 skipped |

The 4 new tests cover: warning on differing metadata, silence on identical metadata (strict mode / HMR), warning when only `appChainIds` differ, and that the stored config really is replaced — so the warning describes what actually happens rather than a guess.

Note `yarn test` is bare `vitest`, so it watches locally; I ran `npx vitest run` from `packages/wallet-sdk` to get a single pass.

### One thing I could not do

`CONTRIBUTING.md` requires GPG-signed commits. There's no GPG key available in the environment I'm working from, and generating one isn't mine to do — so **this commit is unsigned**. Let me know if you need it re-signed and I'll arrange it.

Refs #1860

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01MjiwHkBNkrUKxsQUVeeaSH


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/2</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><code>2</code> if repo is sensitive</td>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>1</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>2</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

